PR #2, opened 2021-08-10
Add initial implementation for common controller and tfstate parsing&conversion

turkenh:
This PR adds an initial implementation (subject to change) for Terraform common controller interacting with the Scheduler interface. 

It also adds support parse tfstate and primitives to do conversions.

I'll open a follow-up pr using this common controller in https://github.com/crossplane-contrib/provider-tf-aws repo.


pkg/reconciler/external.go
```diff
@@ -0,0 +1,127 @@
...
+	}
+
+	dr, err := e.executor.Delete(ctx, tr)
+	if err != nil {
+		return errors.Wrap(err, "failed to delete")
+	}
+	if dr.Completed {
+		return nil
+	}
+	return errors.Errorf("still deleting")
```
negz, line 126: What's going on here? Do we need to wait? 🤔

pkg/reconciler/external.go
```diff
@@ -0,0 +1,127 @@
...
+func (e *TerraformExternal) Update(ctx context.Context, mg xpresource.Managed) (managed.ExternalUpdate, error) {
+	tr, ok := mg.(resource.Terraformed)
+	if !ok {
+		return managed.ExternalUpdate{}, errors.New(errUnexpectedObject)
+	}
+
+	ar, err := e.executor.Apply(ctx, tr)
+	if err != nil {
+		return managed.ExternalUpdate{}, errors.Wrap(err, "failed to apply")
+	}
```
negz, line 106: Do we need to wait for `ar.Completed`?

pkg/conversion/statev4.go
```diff
@@ -0,0 +1,135 @@
+package conversion
+
+import (
+	"encoding/base64"
+	"encoding/json"
+	"github.com/pkg/errors"
+)
+
+const (
+	AnnotationKeyStateMeta = "tf.crossplane.io/state-meta"
```
negz, line 10: Is this file excluded from linting? It seems like it would fail due to missing commentary.

turkenh:
Thanks @negz for the review but this PR is intended to be a draft (just converted) to sync on a high-level structure.
I'll apply your comments for the actual PR if applicable.

turkenh:
Closing in favor of #14

PR #26, opened 2021-08-20
Generate controllers and terraformed interface methods on CRDs

turkenh:
This PR adds 2 new pipelines to our code generation:

- TerraformedGenerator: Generates methods for `Terraformed` interface for each CRD
- ControllerGenerator: Generates controller setup functions for each CRD

See an example usage in: https://github.com/crossplane-contrib/provider-tf-aws/pull/5


pkg/pipeline/controller.go
```diff
@@ -0,0 +1,77 @@
...
+	RootModulePath            string
+	ProviderConfigBuilderPath string
+}
+
+// Generate writes controller setup functions.
+func (tg *ControllerGenerator) Generate(version, kind string) error {
+	groupPkgPath := filepath.Join(tg.RootModulePath, "apis", strings.ToLower(strings.Split(tg.Group, ".")[0]), strings.ToLower(version))
+	kindPkgPath := filepath.Join(groupPkgPath, strings.ToLower(kind))
+
+	pkgPath := filepath.Join(tg.ControllerGroupDir, strings.ToLower(kind))
```
muvaf, line 55:
```suggestion
	pkgDir := filepath.Join(tg.ControllerGroupDir, strings.ToLower(kind))
```

To differentiate it from module path which starts with `github.com`?

pkg/pipeline/controller.go
```diff
@@ -0,0 +1,77 @@
...
+	ctrlFile := wrapper.NewFile(pkgPath, version, templates.ControllerTemplate,
```
muvaf, line 56: `pkgPath` here actually refers to module path rather than the filesystem path.

pkg/pipeline/controller.go
```diff
@@ -0,0 +1,77 @@
...
+		ControllerGroupDir:        controllerGroupDir,
+		RootModulePath:            rootModulePath,
+		ProviderConfigBuilderPath: providerConfigBuilderPath,
+	}
+}
+
+// ControllerGenerator generates controller setup functions.
+type ControllerGenerator struct {
+	Group                     string
+	GroupDir                  string
```
muvaf, line 44:
```suggestion
```

Not used anywhere.

pkg/pipeline/terraformed.go
```diff
@@ -0,0 +1,73 @@
...
+			//  just finding out the identifier field since Terraform uses
+			//  more complex logic like combining multiple fields etc.
+			//  I'll revisit this with
+			//  https://github.com/crossplane-contrib/terrajet/issues/11
+			"Identifier":   terraformID,
+			"ResourceType": terraformName,
+		},
+	}
+
+	pkgPath := filepath.Join(tg.GroupDir, strings.ToLower(kind))
```
muvaf, line 62:
```suggestion
	pkgPath := filepath.Join(cg.RootModulePath, "apis", strings.ToLower(strings.Split(cg.Group, ".")[0]), strings.ToLower(version), strings.ToLower(kind))
```

The path should include version and kind as well since that's the package this file will end up and pkg path `File` expects is literal package path rather than filesystem path.

muvaf (approved):
LGTM! Nothing really blocking.

Issue #20, opened 2021-08-19
Types should be registered


muvaf:
Currently, we [generate](https://github.com/crossplane-contrib/terrajet/pull/4) the CRDs but we also need to add it to the main schema so that it's registered with controller manager cache.

PR #30, opened 2021-08-24
Generate code for registering the schema of the generated types

muvaf:
With this new mechanism, we assume that there is a `var SchemaMap = map[schema.GroupVersion][]runtime.Object` in `apis` folder and all CRDs add themselves to that map instead of the group's own scheme builder. This way we don't have to have a complete list of groups anywhere that we need to iterate through. After the map is populated with `init()`s of types, we add its content to the main schema in `main.go`. This will allow us do some group/type based filtering as well, see https://github.com/crossplane/crossplane/issues/2122 for details.

See https://github.com/crossplane-contrib/provider-tf-aws/pull/6 for example usage of this feature.

Fixes https://github.com/crossplane-contrib/terrajet/issues/20


turkenh (approved):
Looking good. I believe we could fix #31 with a similar approach.

Issue #31, opened 2021-08-24
Controllers setup functions should be registered


turkenh:
### What problem are you facing?

Currently we are [generating controllers setup functions](https://github.com/crossplane-contrib/terrajet/pull/26/files#diff-cf953aeb29e93ecfd0c6a4fc404720111a393dcc8c288bb5d1669443ee11a23fR17) but do not call setup of individual controllers like [this](https://github.com/crossplane/provider-aws/blob/master/pkg/controller/aws.go#L111).

### How could Terrajet help solve your problem?

This sounds like a similar problem to #20 and we might consider introducing a similar mechanism to register controllers.

PR #40, opened 2021-08-27
Change folder structure to old one and add setup & register generators

muvaf:

There have been two collision problems we faced:
* Type name collision between the types of the same CRD.
  * Since Terraform schema uses just `schema.Map` type for all structs, we have to generate a name for the `struct` we generate that holds the fields of that map. We use the field name + `Parameters` or `Observation` depending on where the struct is used. But this causes collisions since there can be fields in two different types with same name, hence two structs generated with the same name.
  * [The solution](https://github.com/crossplane-contrib/terrajet/pull/24) we came up was to prefix the type name with the field names that has been traversed up until that node, i.e. go back and add field names until you end up with a unique name, which is guaranteed as long as you go until the CRD name.
* Type name collisions between the types in the same **group**.
  * Since type names are generated we have a different version of the problem above between the CRDs as well. We initially worked around this problem by putting every CRD into its own package.
  * [The solution](https://github.com/crossplane-contrib/terrajet/pull/24) to the problem above could actually be used for this problem as well. So this PR uses the same approach and reverts the workaround we put in place, i.e. now all CRDs of a single group are in the same package.

Additionally, this PR generates a new file `apis/zz_register.go` and `internal/controller/zz_setup.go` registering API types and controllers. Since the content of these files are dependent on the package structure, I wanted to include them in this PR.

Fixes https://github.com/crossplane-contrib/terrajet/issues/31

I have:

Manually in https://github.com/crossplane-contrib/provider-tf-aws/pull/12


muvaf:
Updated the PR description with more details.

pkg/pipeline/templates/register.go.tmpl
```diff
@@ -0,0 +1,29 @@
...
+	// Register the types with the Scheme so the components can map objects to GroupVersionKinds and back
+	AddToSchemes = append(AddToSchemes,
+		{{- range $alias := .Aliases }}
+		{{ $alias }}SchemeBuilder.AddToScheme,
+		{{- end }}
+	)
+}
+
+// AddToSchemes may be used to add all resources defined in the project to a Scheme
+var AddToSchemes runtime.SchemeBuilder
```
turkenh, line 24: nit: put declaration of `AddToSchemes` on before `init` func which uses it.

pkg/types/builder.go
```diff
@@ -31,34 +31,21 @@ import (
...
+		Package: pkg,
 	}
 }
 
 // Builder is used to generate Go type equivalence of given Terraform schema.
 type Builder struct {
 	Package *types.Package
 
-	genTypes map[string]*types.Named
+	genTypes []*types.Named
```
turkenh, line 42: Any reason to keep this as struct field?

pkg/pipeline/crd.go
```diff
@@ -35,39 +36,40 @@ import (
...
-	// TODO(muvaf): Loading the local package will result in error if there is
-	// any compilation errors, which is the case before running kubebuilder
-	// generators. For now, we act like the target package is empty.
-	pkg := types.NewPackage(kindPkgPath, strings.ToLower(version))
-	file := wrapper.NewFile(pkg.Path(), pkg.Name(), templates.CRDTypesTemplate,
+	file := wrapper.NewFile(cg.pkg.Path(), cg.pkg.Name(), templates.CRDTypesTemplate,
 		wrapper.WithGenStatement(GenStatement),
 		wrapper.WithHeaderPath("hack/boilerplate.go.txt"), // todo
-		wrapper.LinterEnabled(),
+		// wrapper.LinterEnabled(),
```
turkenh, line 61: Is this comment out intentional?

turkenh: there are other instances commented out as well in other files

pkg/pipeline/setup.go
```diff
@@ -0,0 +1,62 @@
...
+
+// NewSetupGenerator returns a new SetupGenerator.
+func NewSetupGenerator(rootDir, modulePath string) *SetupGenerator {
+	return &SetupGenerator{
+		LocalDirectoryPath: filepath.Join(rootDir, "internal", "controller"),
+		ModulePath:         modulePath,
+	}
+}
+
+// SetupGenerator generates scheme registration file.
```
turkenh, line 38: ...generates setup file

pkg/pipeline/templates/controller.go.tmpl
```diff
@@ -14,6 +15,6 @@ import (
 )
 
 // Setup adds a controller that reconciles {{ .CRD.Kind }} managed resources.
-func Setup(mgr ctrl.Manager, l logging.Logger) error {
+func Setup(mgr ctrl.Manager, l logging.Logger, _ workqueue.RateLimiter) error {
```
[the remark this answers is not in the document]
muvaf: The rate limiter currently is configured in `main.go` of providers. I thought it's not possible to use it for some reason but it seems it's just a controller-runtime option we can pass through as you suggested.

pkg/pipeline/templates/register.go.tmpl
muvaf: Just obeying the existing template here https://github.com/crossplane/provider-template/blob/master/apis/template.go#L27

pkg/types/builder.go
muvaf: In every node of the recursion tree, the resulting types are added to that array. So, we need to be able to access the same array in each iteration of `buildResource`.

pkg/pipeline/crd.go
muvaf: Just removed all. It turns out running `goimports` once for all files take much less time compared to having it run for every file before writing them to disk.

pkg/terraform/controller.go
```diff
@@ -3,6 +3,8 @@ package terraform
 import (
 	"context"
 
+	"k8s.io/client-go/util/workqueue"
```
turkenh, line 6: nit: goimports

turkenh (approved):

muvaf:
@turkenh I made an addition in the commit I pushed after your approval. It removes the generic `Setup` function and lets managed resource `controller.go` to call `NewConnector` directly since there isn't any changes we make in that generic function now that the rate limiter is taken from main.

turkenh:
> @turkenh I made an addition in the commit I pushed after your approval. It removes the generic `Setup` function and lets managed resource `controller.go` to call `NewConnector` directly since there isn't any changes we make in that generic function now that the rate limiter is taken from main.

Sounds good to me @muvaf.

PR #50, opened 2021-09-04
Add comments builder with marker parsing option

### LINKED ISSUE #8: Comment tags on CRD fields
Generated CRDs don't have some of the necessary tags like `// +kubebuilder:validation:Required`. A comment printer functionality will be necessary and it will help with printing documentation as well.

### LINKED ISSUE #33: Handle provider configuration differences with Terraform

There are some differences between Crossplane and Terraform on which fields are passed as provider configuration vs resource specification. A good example of this is "region" field for AWS, which is part of provider configuration on Terraform side but part of resource spec for Crossplane.

We need to discuss how to handle these differences and presumably follow Crossplane's conventions which would require some changes during schema generation and conversion side.

turkenh:

This PR introduces a way to provide terrajet configuration with the help of comment markers which is required to resolve #33 as part of https://github.com/crossplane-contrib/provider-tf-aws/pull/10. It also introduces a comments builder utility fixing #8.

Please note, this PR is currently on top of #40 to prevent possible conflicts.

Fixes #8

I have:

Unit tests
With provider-tf-aws change in [this commit](https://github.com/crossplane-contrib/provider-tf-aws/pull/10/commits/cfcdfa33d1e3bc2c0ac3b4d93b05abb4f2e8c394).


pkg/comments/builder.go
```diff
@@ -0,0 +1,61 @@
...
+	"github.com/crossplane-contrib/terrajet/pkg/markers"
+)
+
+const (
+	errCannotParseAsFieldMarker = "cannot parse line as marker for field"
+)
+
+// Builder builds comments for fields by also building terrajet options
+// supported by comment markers
+type Builder struct {
```
muvaf, line 17: It seems that we're utilizing `controller-tools` marker parsing mechanisms but do you think we need this tool to be that generic given that it lives in `terrajet` repo, not `crossplane-tools`? What do you think about having fields for each possible marker `terrajet` knows, like `Required bool`, and use that as medium for `Builder` here? Very similar to current design but instead of only `CRDTags`, there would be all other options we know (required, minimum, reference etc.) as fields in `Options` and developers would just need to worry about this single `Builder` struct that has an `Options` field. The `Builder` would only parse/build the options it knows of and unrecognized lines would be treated just like any other comment text.

It's a rough general idea, I'm just wondering if we can leverage the fact that we're in a specific repo to make the operation more strong-typed and easy to work with since we don't have to pay the cost of being so generic like `crossplane-tools` or `controller-tools`.

turkenh: Actually, I am not trying to be generic here, rather trying to keep things scoped. This `comments builder` is responsible for building added comment lines as proper comments, e.g. Regular comments listed before marker ones. One exception here, which I am also not much happy about is, it can also build terrajet marker options if input comments have markers that terrajet recognizes. The `markers` package on the other hand is responsible for handling terrajet recognized markers and means to build kubebuilder markers with strongly typed structs.

Regarding kubebuilder options, I also considered having possible configurations in a single struct, but after checking supported kubebuilder markers (e.g. for [validation](https://book.kubebuilder.io/reference/markers/crd-validation.html)), I didn't prefer a struct growing with all possible parameters there. I believe the current approach is straightforward enough with the helpers added (e.g. `markers.MarkerForConfig()`) but still extendible especially considering there could be some markers that require (multiple) parameters. 

If we really want a struct with support options, I would rather prefer to build another layer on top which uses the above mechanisms, but I am hesitant that it could grow very quickly with a lot of options.

muvaf: > Regarding kubebuilder options, I also considered having possible configurations in a single struct, but after checking supported kubebuilder markers (e.g. for validation), I didn't prefer a struct growing with all possible parameters there.


If we want to cover all kubebuilder markers, yes, but I'm not sure that's where we'll end up. The use cases I'm able to come up for this comment utility:
* Let users add markers Terrajet recognizes, like overriding a field tag.
* Let Terrajet read markers it can act upon, like overriding a field tag. 
* Let Terrajet add markers for certain validations derived from terraform.Schema, like kubebuilder `required` marker.

In all cases, we actually have enough information to scope the options. For example, the set of `kubebuilder` tags we'll ever need is limited by what [TF Schema](https://github.com/hashicorp/terraform-plugin-sdk/blob/9321fe1/helper/schema/schema.go#L37) exposes, which seems like `required`, min/max, and possibly input default. The rest isn't really needed to be recognized by Terrajet because it won't take any action anyway. For other two use cases, Terrajet is either writer or reader of the same information.

If we have all this in a strong-typed medium, we can use that type both as internal representation of a comment in Terrajet and also exception input for specific fields from the author of provider.

turkenh: @muvaf refactored for a simpler interface considering the limited requirements we have here, as we also discussed offline. I dropped support for multiple keys in a single marker to simplify further (completely removing dependency to controller-tools etc.), and we can consider adding support for that when/if we need that functionality.

pkg/comments/comment.go
```diff
@@ -0,0 +1,78 @@
...
+// New returns a Comment by parsing Terrajet markers as Options
+func New(text string, opts ...Option) (*Comment, error) {
+	to := markers.TerrajetOptions{}
+
+	var lines []string
+	for _, line := range strings.Split(text, "\n") {
+		line = strings.TrimSpace(line)
+		if line == "" {
+			continue
+		}
```
muvaf, line 35: Empty line could be valid for multi-paragraph comments.

pkg/comments/comment.go
```diff
@@ -0,0 +1,78 @@
...
+func (c *Comment) String() string {
+	return c.Text + "\n" + c.Options.String()
+}
+
+// Build builds comments by adding comment prefix ("// ") to each line of
+// the string representation of this Comment.
+func (c *Comment) Build() string {
+	all := strings.ReplaceAll("// "+c.String(), "\n", "\n// ")
+	return strings.TrimSuffix(all, "// ")
+}
```
muvaf, line 78:
```suggestion
// String builds comments by adding comment prefix ("// ") to each line of
// the string representation of this Comment.
func (c *Comment) String() string {
	all := strings.ReplaceAll("// "+c.String(), "\n", "\n// ")
	return strings.TrimSuffix(all, "// ")
}
```

nitpick: is there a case where the string representation of the comment is useful without `//`?

pkg/markers/terrajet.go
```diff
@@ -0,0 +1,59 @@
...
+
+const (
+	markerPrefixTerrajet = "+terrajet:"
+
+	errFmtCannotParse = "cannot parse as a terrajet prefix: %s"
+)
+
+var (
+	markerPrefixCRDTFTag   = fmt.Sprintf("%scrdfield:TFTag=", markerPrefixTerrajet)
+	markerPrefixCRDJsonTag = fmt.Sprintf("%scrdfield:JsonTag=", markerPrefixTerrajet)
```
muvaf, line 18:
```suggestion
	markerPrefixCRDJsonTag = fmt.Sprintf("%scrdfield:JSONTag=", markerPrefixTerrajet)
```

pkg/markers/terrajet.go
```diff
@@ -0,0 +1,59 @@
...
+)
+
+const (
+	markerPrefixTerrajet = "+terrajet:"
+
+	errFmtCannotParse = "cannot parse as a terrajet prefix: %s"
+)
+
+var (
+	markerPrefixCRDTFTag   = fmt.Sprintf("%scrdfield:TFTag=", markerPrefixTerrajet)
```
muvaf, line 17:
```suggestion
	markerPrefixCRDTFTag   = fmt.Sprintf("%scrd:field:TFTag=", markerPrefixTerrajet)
```

Would `+terrajet:crd:field:TFTag` be more open to expansions in the future, like `+terrajet:crd:type:NameOverride`?

pkg/markers/crossplane.go
```diff
@@ -0,0 +1,19 @@
...
+
+import "fmt"
+
+// CrossplaneOptions represents the Crossplane marker options that terrajet
+// would need to interact
+type CrossplaneOptions struct {
+	ReferenceToType string
+}
+
+func (o CrossplaneOptions) String() string {
```
muvaf, line 11: Similar comment with `Build()` vs `String()`. Having `String()` print the latest `//`-prefixed string looks easier to know which one users should interact with, but no strong opinion either way.

go.mod
```diff
@@ -3,15 +3,17 @@ module github.com/crossplane-contrib/terrajet
 go 1.16
 
 require (
-	github.com/crossplane/crossplane-runtime v0.14.0
+	github.com/crossplane/crossplane-runtime v0.15.0
+	github.com/google/go-cmp v0.5.6
+	github.com/hashicorp/hcl/v2 v2.8.2 // indirect
```
muvaf, line 8: Is this dependency necessary? I removed it and `make reviewable` still worked.

muvaf (approved):
LGTM! Thank you so much @turkenh for the quick turnaround on this! I have a couple of comments but mostly nitpicks, nothing blocking.

pkg/comments/builder.go
muvaf: @turkenh Thank you!

pkg/comments/comment.go
turkenh: Yes, it is used while building a comment with terrajet markers here: https://github.com/crossplane-contrib/provider-tf-aws/pull/10/commits/cfcdfa33d1e3bc2c0ac3b4d93b05abb4f2e8c394#diff-f59d142a3ad45f7a7bb182ca95d3d7130496d5269597d8217bb1ec16b63295cbR160

We parse that back in [here](https://github.com/crossplane-contrib/terrajet/pull/50/files#diff-f582aaab14f02ada1f64e64e80337f2a797b8e38ffe5c3288b38c98e67824a86R67). Always printing `//` would require us to be able to consider that while parsing.

pkg/markers/terrajet.go
turkenh: We could change it later since it is no longer part of our API (private parameter of the package), but doing it now also sounds good.

pkg/markers/crossplane.go
turkenh: `comments.Comment` is an object that you could build with a text (e.g. description) and some options. `String()` returns its string representation where all options are represented as markers. `Build()` on the other hand builds it as a comment to print. Still, I would agree for a simplification if we didn't have a use case where we also need to parse input text for markers, where using `String()` makes more sense, please see https://github.com/crossplane-contrib/terrajet/pull/50#discussion_r706215430

go.mod
turkenh: Removed, I am not sure how it ended there but `go mod tidy` does not remove it either.

PR #49, opened 2021-09-03
Add .CRD.Category template param for crd_types.go.tmpl

ulucinar:
- Generated CRDs have the provider name as a category

Signed-off-by: Alper Rifat Ulucinar <ulucinar@users.noreply.github.com>

### Description of your changes

Fixes #

I have:

- [x] Read and followed Crossplane's [contribution process].
- [x] Run `make reviewable` to ensure this PR is ready for review.
- [ ] Added `backport release-x.y` labels to auto-backport this PR if necessary.

### How has this code been tested

[contribution process]: https://git.io/fj2m9

PR #52, opened 2021-09-08
Add late-initialization for Terraformed resources

### LINKED ISSUE #10: Implement LateInitialization
In Crossplane managed resources, not all fields in the spec are required during creation time.
For optional fields, we need to properly parse attributes in Terraform state into generated spec schema.

ulucinar:

Introduces late-initialization for `Terraformed` resources. Copies late-initialization library from provider-aws `pkg/controller/cloudfront/` package. 

Fixes #10

I have:

I have manually tested this PR with [provider-tf-azure](https://github.com/crossplane-contrib/provider-tf-azure/pull/1). 
When a `VirtualNetwork` MR with the following manifest:
```yaml
apiVersion: virtual.azure.tf.crossplane.io/v1alpha1
kind: VirtualNetwork
metadata:
  name: example
spec:
  deletionPolicy: Orphan
  forProvider:
    addressSpace:
    - 10.0.0.0/16
    dnsServers:
    - 10.0.0.1
    - 10.0.0.2
    - 10.0.0.3
    location: East US
    name: importDemo
    resourceGroupName: alper
    tags:
      purpose: demo
      key1: value1
  providerConfigRef:
    name: example
```
is provisioned and then deleted, and after a new `VirtualNetwork` MR is imported using the following manifest:
```yaml
apiVersion: virtual.azure.tf.crossplane.io/v1alpha1
kind: VirtualNetwork
metadata:
  name: imported
  annotations:
    crossplane.io/external-name: /subscriptions/.../resourceGroups/alper/providers/Microsoft.Network/virtualNetworks/importDemo
spec:
  forProvider:
    name: "importDemo"
    location: "East US"
    resourceGroupName: "alper"
    addressSpace:
      - "10.0.0.0/16"
  providerConfigRef:
    name: example
```
, the imported MR is late-initialized as follows:
```yaml
apiVersion: virtual.azure.tf.crossplane.io/v1alpha1
kind: VirtualNetwork
metadata:
  annotations:
    crossplane.io/external-name: /subscriptions/038f2b7c-3265-43b8-8624-c9ad5da610a8/resourceGroups/alper/providers/Microsoft.Network/virtualNetworks/importDemo
    creationTimestamp: "2021-09-10T03:18:29Z"
...
  finalizers:
  - finalizer.managedresource.crossplane.io
  name: imported
...
spec:
  deletionPolicy: Delete
  forProvider:
    addressSpace:
    - 10.0.0.0/16
    dnsServers:
    - 10.0.0.1
    - 10.0.0.2
    - 10.0.0.3
    location: East US
    name: importDemo
    resourceGroupName: alper
    tags:
      key1: value1
      purpose: demo
  providerConfigRef:
    name: example
status:
  atProvider:
    guid: 1222d5b8-41b7-4d84-bca0-a8282390422e
  conditions:
  - lastTransitionTime: "2021-09-10T03:18:39Z"
    reason: Available
    status: "True"
    type: Ready
  - lastTransitionTime: "2021-09-10T03:18:39Z"
    reason: ReconcileSuccess
    status: "True"
    type: Synced
```


ulucinar:
This PR is on top of on #3 and #49. 

It looks like Terraform state file (`terraform.tfstate`) contains zero values of the _element types_ (of optional parameters) instead of a JSON `null` value. For example, in the schema of an `azurerm_virtual_network`, the `bgp_community` parameter is an optional one and for a `VirtualNetwork` for which `bgp_community` is not set, here is the state file:

```json
{"version":4,"terraform_version":"1.0.3","serial":4,"lineage":"d532ae54-6f57-b401-b7a4-aa44841fe45f","outputs":{},"resources":[{"mode":"managed","type":"azurerm_virtual_network","name":"example","provider":"provider[\"registry.terraform.io/hashicorp/azurerm\"]","instances":[{"schema_version":0,"attributes":{
            "address_space": [
              "10.0.0.0/16"
            ],
            "bgp_community": "",
            "ddos_protection_plan": [],
            "dns_servers": [],
            "guid": "14e18c78-8dd0-4e9e-8e11-eea1810bc6d0",
            "id": "/subscriptions/.../resourceGroups/alper/providers/Microsoft.Network/virtualNetworks/virtualNetwork1",
            "location": "eastus",
            "name": "virtualNetwork1",
            "resource_group_name": "alper",
            "subnet": [],
            "tags": {},
            "timeouts": null,
            "vm_protection_enabled": false
          },
```

And here is the corresponding late-initialized MR:

```yaml
apiVersion: virtual.azure.tf.crossplane.io/v1alpha1
kind: VirtualNetwork
metadata:
  name: example
  ...
spec:
  deletionPolicy: Delete
  forProvider:
    addressSpace:
    - 10.0.0.0/16
    bgpCommunity: ""
    location: East US
    name: virtualNetwork1
    resourceGroupName: alper
    vmProtectionEnabled: false
  providerConfigRef:
    name: example
status:
...
```

And `""` is not a valid value for `bgp_community`, so it should not be a default set for it.

Although the `bgp_community` parameter is not set for the resource, Terraform state has a zero value for its element type (empty string) and it gets late-initialized because its value is not `nil` for the corresponding Go type of `*string`. Will take a look at why it's serialized as an empty string.

Interestingly, for certain fields (probably those are specially/explicitly treated fields that are **not** resource specific in Terraform code, I did not check yet), we observe `null` values. An example is `.resources[0].instances[0].attributes.timeouts`.

ulucinar:
Probably it makes sense to _approximate_ late initialization by treating zero-valued attributes as unset, and generate late-initialization code per MR to allow late-initialization logic to be overridden for certain/desired resources, where the generic logic is not appropriate.

muvaf (commented):
Most of my comments are because of the change in the context of where this functionality was used, i.e. it was in `provider-aws` written in a way that would be very generic but here we have some assumptions we can safely make.

There are a few drawbacks with `reflect` package in general that is not specific to this PR:
* It can [`panic`](https://thanos.io/tip/contributing/coding-style-guide.md/#never-use-panics) unexpectedly, which is OK during code generation but not really in runtime.
  * Handling errors via `panic` is [not idiomatic](https://github.com/golang/go/wiki/CodeReviewComments#dont-panic).
* We cannot really wrap the errors with understandable strings while bubbling when a `panic` occurs.
* I haven't run any tests, but `reflect` is [said](https://thanos.io/tip/contributing/coding-style-guide.md/#avoid-using-the-reflect-or-unsafe-packages) to be slow.

We agreed on using this utility we already had to unblock many stuff but in my opinion, using `reflect` for a runtime operation that is on the hot execution path is not the final state we'd like to be in in the future. We can merge this PR after addressing the comments but I'd suggest opening a tracking issue to remove `reflect`. I think that we can have this functionality implemented in two ways in the future:
* Work with `map[string]interface{}` instead of reflection of `interface{}`.
  * This could be a single function that is used by all resources just like the implementation in this PR but it'd work on the object itself, not with its reflection, which removes most of the concerns with `reflect`.
  * In fact, the recursion logic wouldn't change much.
* Generate strong-typed `LateInitialize` functions.
  * This option is the best and can be used in ALL providers (a comment marker to enable generation of a `LateInitialize(a, b *SomeParametres) bool`) but it has high cost due to required development effort if we do it from the scratch. There are utilities in [typewriter](https://github.com/muvaf/typewriter#quick-start) that could allow this to be implemented quickly in `crossplane-tools` possibly, but we decided not to use it for complex stuff we can't easily change for now.

Alternatively, we can make this PR work on `map[string]interface{}` since the recursion logic would be very similar.

muvaf:
> Work with map[string]interface{} instead of reflection of interface{}. 

We synced with @ulucinar and he made me realize that even if we use `map[string]interface{}` we'd still have to use reflection, so it's not really worth it. We might revisit to see if we should generate strong typed functions if we see performance drop in our tests when we use reflection, but right now using `reflect` sounds like the best decision.

turkenh:
> There are a few drawbacks with reflect package in general that is not specific to this PR:

Just noticed that [jsoniter](https://github.com/json-iterator/go) using https://github.com/modern-go/reflect2 to avoid performance drawbacks of the reflect library, worth considering here if performance is a real concern.

ulucinar:
> > There are a few drawbacks with reflect package in general that is not specific to this PR:
> 
> Just noticed that [jsoniter](https://github.com/json-iterator/go) using https://github.com/modern-go/reflect2 to avoid performance drawbacks of the reflect library, worth considering here if performance is a real concern.

Hi @turkenh,
Thanks for bringing that library up. It's definitely worth looking.

muvaf (approved):
LGTM! @ulucinar Thank you for taking care of this! It's always a challenge to work with recursion.

ulucinar:
Thanks @muvaf. Unlike the [original library](https://github.com/crossplane/provider-aws/blob/master/pkg/controller/cloudfront/lateinit.go), the proposed change here can now only handle same type structs as sources and destinations, as in the Terrajet context it does not make sense to support late-initializing between different types. This greatly simplifies the PR. I have also removed handling certain edge cases as suggested. If we need them in the future, we may have to add them back.

Regarding the performance of reflection, I will do some experiments & measurements and as @turkenh suggested above, we may consider switching from std reflection library to https://github.com/modern-go/reflect2, or implementing the proposed logic in this PR in code generation.

Issue #15, opened 2021-08-18
Add known special acronyms to init


muvaf:
The library we use to do camel case conversions [allow](https://github.com/iancoleman/strcase/#custom-acronyms-for-tocamel--tolowercamel) customization by adding acronyms in `init()`. We should have a default set of custom acronyms (like DNS, IPv6, ID etc) and let providers add their own if need be.

muvaf:
Ugh, it turns out it [doesn't run](https://github.com/iancoleman/strcase/blob/692d1b8/camel.go#L38) the checker for every word in the string; it just acts like an override map. I'll have to implement a converter for `snake_field to CamelCase`.

muvaf:
Hmm Kubernetes resources do appear to care about this, there is `status.podIP` field in `Pod`.

negz:
> I'm having some second thoughts about whether we should do this for json tags

The goal is to match the cloud APIs as closely as possible, so I actually think not doing this for JSON tags might make sense in a lot of cases. For example I believe AWS APIs typically use JSON struct fields like `vpcId` or `someArn`, not `vpcID` and `someARN`. I thiiiiink GCP is like this too - I just checked https://cloud.google.com/sql/docs/mysql/admin-api/rest/v1/instances#DatabaseInstance and I see `secondaryGceZone` not `secondaryGCEZone`.

It would still be nice to correctly capitalise the actual Go struct fields in our types, in accordance with typical Go style. I could be convinced it's not the end of the world if it's not possible to do that though - there's plenty of precedent for relaxing style on generated code that humans don't need to interface with as often.

Issue #27, opened 2021-08-22
Some nested status fields might be missing


muvaf:
### What problem are you facing?

We use Terraform schema to generate the type of every field. While Terraform has one big untyped tree of fields to hold all the information, that's not the case in Crossplane. In line with Kubernetes API conventions, we separate fields into `spec` and `status` fields, `status` ones being fields that user cannot ever configure.

In Terraform schema, fields report whether they are `Computed` and `Optional`. The fields that are both `Computed` and `Optional` are the ones that have server-side defaults and user-configurable; the ones we late-initialize. So we [put those under spec](https://github.com/crossplane-contrib/terrajet/blob/df8f6cf/pkg/types/builder.go#L87), but fields that are `Computed` and `not Optional` [go to `status`](https://github.com/crossplane-contrib/terrajet/blob/df8f6cf/pkg/types/builder.go#L168).

The problem is that [there can be](https://github.com/crossplane-contrib/terrajet/blob/df8f6cf/pkg/types/builder.go#L155) some types that are `Computed` and `not Optional` but also has nested fields that can go under `spec` and vice versa. Right now, when we compute the type for a field, we return a parameters type as well as an observation type, then decide which one to use depending on the properties of the field. If the field goes to spec but there are nested fields that could go to status, they are eliminated. Also, if a field goes to status, all nested fields go to status as well and the parameter types are ignored.

When I did an analysis of how prevalent this problem is in, for example, AWS, I see the following output:
```
there are 1 fields that could be observation type of .SyntheticsCanary.VpcConfig but ignored since the field is a parameters field
there are 2 fields that could be observation type of .AmplifyDomainAssociation.SubDomain but ignored since the field is a parameters field
there are 1 fields that could be observation type of .StoragegatewayGateway.SmbActiveDirectorySettings but ignored since the field is a parameters field
there are 1 fields that could be observation type of .LexBotAlias.ConversationLogs.LogSettings but ignored since the field is a parameters field
there are 1 fields that could be observation type of .Route53ResolverEndpoint.IpAddress but ignored since the field is a parameters field
there are 1 fields that could be observation type of .LambdaFunction.VpcConfig but ignored since the field is a parameters field
there are 1 fields that could be observation type of .MwaaEnvironment.LoggingConfiguration.DagProcessingLogs but ignored since the field is a parameters field
there are 1 fields that could be observation type of .MwaaEnvironment.LoggingConfiguration.SchedulerLogs but ignored since the field is a parameters field
[...]

PR #53, opened 2021-09-08
Handling camel case conversions

muvaf:

I was expecting that only thing needed would be the list of known acronyms but it turns out the library we chose doesn't check its acronym map for every word; it checks the full string in the beginning and that's it. But that doesn't cover most of our cases, for example it doesn't make `id -> ID` in `subnetId`. It converts `ID` to `id` only if the whole string is just `ID`. In fact, it doesn't convert from lower to upper at all; the [map holds](https://github.com/iancoleman/strcase/blob/master/acronyms.go#L4) upper case as key, hence it doesn't work for us at all because all of our input is snake field.

So, I added a wrapper that will call `strcase.ToCamel` for each word if there isn't a match in our local map and it will do this operation only once, producing a `Name` object that holds all variants of the name we need.

Fixes https://github.com/crossplane-contrib/terrajet/issues/15

I have:

See https://github.com/crossplane-contrib/provider-tf-aws/pull/13


pkg/types/name.go
```diff
@@ -0,0 +1,121 @@
...
+var (
+	// Used to hold lower -> camel known exceptions.
+	lowerToCamelAcronyms = map[string]string{
+		"id": "ID",
+	}
+)
+
+// AddAcronym is used to add exception words that will be used to intervene
+// the conversion from lower case to camel case.
+func AddAcronym(lower, camel string) {
```
ulucinar, line 67: Should we mention in the doc comment that the intended use case is to call this in an `init()` function as a one-time configuration (not safe for concurrent use)? 

nit: While I see that `strcase` has employed a global variable to hold acronym mapping, we may also consider moving to `NameGenerator` approach that holds its own configuration. But definitely not a blocker from my perspective.

pkg/types/name.go
```diff
@@ -0,0 +1,121 @@
...
+// For provider-specific ones, like ARN for Amazon Web Services, provider
+// authors need to configure them in their provider repository.
+
+// NOTE(muvaf): We can have more maps like camel -> lower for reverse conversion,
+// but it's not necessary for now.
+
+var (
+	// Used to hold lower -> camel known exceptions.
+	lowerToCamelAcronyms = map[string]string{
+		"id": "ID",
```
ulucinar, line 61: Should we move this to`init()` or vice versa?

pkg/types/builder.go
```diff
@@ -61,13 +60,12 @@ func (g *Builder) buildResource(res *schema.Resource, names ...string) (*types.N
...
 	for _, snakeFieldName := range keys {
 		sch := res.Schema[snakeFieldName]
-		fieldName := strcase.ToCamel(snakeFieldName)
-		lowerCamelFieldName := strcase.ToLowerCamel(snakeFieldName)
-		fieldType, err := g.buildSchema(sch, append(names, fieldName))
+		fieldName := NewNameFromSnake(snakeFieldName)
+		fieldType, err := g.buildSchema(sch, append(names, fieldName.Camel))
 		if err != nil {
-			return nil, nil, errors.Wrapf(err, "cannot infer type from schema of field %s", fieldName)
+			return nil, nil, errors.Wrapf(err, "cannot infer type from schema of field %s", fieldName.Snake)
```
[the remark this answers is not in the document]
muvaf: No, we calculate the type name at the end of this function only if all succeeds and it doesn't always match a familiar name you derive from Terraform schema because of the uniqueness problem there. And if an error is returned here, we don't print the type hence users would see an error with a type name that doesn't exist because Terraform schema doesn't name types; they are all `map`s with certain keys. So, the only information that is most useful for them at this point is the snake name of the field that they can use to debug the TF schema.

pkg/types/builder.go
```diff
@@ -76,12 +74,12 @@ func (g *Builder) buildResource(res *schema.Resource, names ...string) (*types.N
 		switch {
 		case sch.Computed && !sch.Optional:
 			obsFields = append(obsFields, field)
-			obsTags = append(obsTags, fmt.Sprintf("json:\"%s\" tf:\"%s\"", lowerCamelFieldName, snakeFieldName))
+			obsTags = append(obsTags, fmt.Sprintf("json:\"%s\" tf:\"%s\"", fieldName.LowerCamel, fieldName.Snake))
 		default:
 			if sch.Optional {
-				paramTags = append(paramTags, fmt.Sprintf("json:\"%s,omitempty\" tf:\"%s\"", lowerCamelFieldName, snakeFieldName))
+				paramTags = append(paramTags, fmt.Sprintf("json:\"%s,omitempty\" tf:\"%s\"", fieldName.LowerCamel, fieldName.Snake))
```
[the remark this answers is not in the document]
muvaf: I don't really have a strong opinion about it but it might have some side effects that are not related to this PR. So I'd prefer changing that in a separate PR, possibly in your late init PR where you can experiment with it?

pkg/types/builder.go
```diff
@@ -76,12 +74,12 @@ func (g *Builder) buildResource(res *schema.Resource, names ...string) (*types.N
 		switch {
 		case sch.Computed && !sch.Optional:
```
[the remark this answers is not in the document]
muvaf: There is some discussion [here](https://github.com/crossplane-contrib/terrajet/issues/27).

The reason we're putting only those into status is that the fields that are computed but also optional are the ones that are defaulted by API server. So, users can give an input there but if they don't, server decides on a value and sends it back, which is then late-initialized by us. So we can't really put those fields under status since users are able to enter input there.

pkg/types/name.go
muvaf: I'll add a comment about the intention for where it should be called from. I wouldn't really prefer the name generator approach though because acronym sets do not change in the runtime at all due to their nature, so having this as global variable that `NewNameFromSnake` can use looks simpler to me.

pkg/types/name.go
muvaf: I added that just as an example pairing so it's clear that the key is lower case and value is camel but I can move it to `init()` if you have a strong feeling.

ulucinar (approved):

Issue #11, opened 2021-08-13
Setting external-name annotation


turkenh:
Crossplane uses external-name annotation to uniquely identify external resource (typically living in cloud API): https://crossplane.io/docs/v0.11/introduction/managed-resources.html#external-name

Terraform has a similar concept, [ID](https://www.terraform.io/docs/glossary.html#id), which is used for the same purpose. 
However, terraform resources do not use the same field in attributes as an id rather this is a piece of information documented per resource. See https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/route53_zone#import

From [import documentation](https://www.terraform.io/docs/cli/commands/import.html#usage):

> ID is dependent on the resource type being imported. For example, for AWS instances it is the instance ID (i-abcd1234) but for AWS Route53 zones it is the zone ID (Z12ABC4UGMOZ2N). Please reference the provider documentation for details on the ID format. If you’re unsure, feel free to just try an ID. If the ID is invalid, you’ll just receive an error message.

We need to figure out how can we auto-generate this information and use it while setting external name annotation.

muvaf:
@turkenh Right now, we assume that `.id` is the field to use in state and it seems that's what's used for `terraform import` as well. The only thing missing is that Terraform does not guarantee that the field will be named as `id` so we might need to override that default. So, I guess this issue can be closed if we provide a way to override the path to ID field in state, right?

muvaf:
Oh actually we don't optimize for resources that can be named, for example [GCP Network](https://registry.terraform.io/providers/hashicorp/google/latest/docs/resources/compute_network). If users do give an external-name, we don't use that for creation of the resource from what I can tell, right? They have to give the name in `spec.forProvider.name`?

muvaf:
I just tested that giving `id` attribute in the input HCL/JSON is not enough for Terraform to figure out the rest of the fields during the creation. The following JSON returns error saying `name` is required:
```json
{
  "terraform": {
      "required_providers": {
          "google": {
              "source": "hashicorp/google",
              "version": "3.5.0"
            }
        }
    },
  "provider": {
    "google": [
      {
        "project": "some-project",
        "region": "us-central1",
        "zone": "us-central1-c"
      }
    ]
  },
  "resource": {
    "google_compute_network": {
      "vpc_network": {
        "id": "projects/some-project/global/networks/muvaf-testing-network"
      }
    }
  }
}
```
```
$ terraform apply
╷
│ Error: Missing required argument
│
[...]

Issue #56, opened 2021-09-09
Generate reference and selector fields if reference marker exists


muvaf:
### What problem are you facing?

https://github.com/crossplane/crossplane-tools/pull/35 lets us generate reference resolver but it still needs the `Reference` and `Selector` fields generated beforehand.

### How could Terrajet help solve your problem?

Make schema buidler add those two fields each time it sees the comment marker resolver generator uses.

PR #66, opened 2021-09-15
External name to parameters mapping

muvaf:

First stab at opening up an interface for per-resource configuration and external name function is the first implementation. As described in https://github.com/crossplane-contrib/terrajet/issues/11#issuecomment-919154858 we're going to ask authors to provide a set of fields to omit from top level spec and give a function that will take external name and set the necessary fields in the output HCL before actual parameters are applied.

Fixes https://github.com/crossplane-contrib/terrajet/issues/11

I have:

Code compiles with provider-tf-aws but I haven't tried provisioning a resource. See https://github.com/crossplane-contrib/provider-tf-aws/pull/19

### DIFF
--- a/pkg/terraform/resource/configuration.go
+++ b/pkg/terraform/resource/configuration.go
@@ -67,6 +67,13 @@ type ExternalNamer struct {
 	// they are specified via external name. You can omit only the top level fields.
 	// No field is omitted by default.
 	OmittedFields []string
+
+	// DisableNameInitializer allows you to specify whether the name initializer
+	// that sets external name to metadata.name if none specified should be disabled.
+	// It needs to be disabled for resources whose external name includes information
+	// more than the actual name of the resource, like subscription ID or region
+	// etc. which is unlikely to be included in metadata.name
+	DisableNameInitializer bool
 }
 
 // Configuration is the set of information that you can override at different steps
--- a/pkg/terraform/resource/configuration.go
+++ b/pkg/terraform/resource/configuration.go
@@ -27,10 +27,11 @@ func NopConfigureWithName(_ map[string]interface{}, _ string) {}
 // ConfigurationOption allows setting optional fields of a Configuration object.
 type ConfigurationOption func(*Configuration)
 
-// WithExternalName allows you to set an ExternalNamer for given Configuration.
-func WithExternalName(e ExternalNamer) ConfigurationOption {
+// WithExternalNameConfiguration allows you to set an ExternalNameConfiguration
+// for given Configuration.
+func WithExternalNameConfiguration(e ExternalNameConfiguration) ConfigurationOption {
 	return func(c *Configuration) {
-		c.ExternalNamer = e
+		c.ExternalName = e
 	}
 }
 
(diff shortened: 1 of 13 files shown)


muvaf:
I investigated ways to avoid getting a type path but it doesn't seem possible. Hence opted for having a `SelfVarPath` in the struct that the users will give so that we can have some signature checking in compilation time since the func is part of that struct now and the path string of `var` will be close to the `var` itself.

pkg/terraform/resource/configuration.go
```diff
@@ -0,0 +1,99 @@
...
+
+// ConfigureWithNameFn functions configure the base resource fields by using
+// given name value.
+type ConfigureWithNameFn func(base map[string]interface{}, name string)
+
+// ConfigurationOption allows setting optional fields of a Configuration object.
+type ConfigurationOption func(*Configuration)
+
+// WithExternalName allows you to set an ExternalNamer for given Configuration.
+func WithExternalName(e ExternalNamer) ConfigurationOption {
```
turkenh, line 27:
```suggestion
func WithExternalNamer(e ExternalNamer) ConfigurationOption {
```

pkg/terraform/resource/configuration.go
```diff
@@ -0,0 +1,99 @@
...
+	for _, f := range opts {
+		f(c)
+	}
+	return c
+}
+
+// ExternalNamer contains all information that is necessary for naming operations,
+// such as removal of those fields from spec schema and calling Configure function
+// to fill attributes with information given in external name.
+type ExternalNamer struct {
```
turkenh, line 57: `ExternalNamer` sounds like it contains configuration/information to build the external name (using existing fields), whereas, it is mostly for almost the opposite, to parse the external name into spec. Could we find a more appropriate name for this type to prevent possible confusion, e.g. `NamingConfiguration` ?

pkg/tfcli/templates/main.tf.json.tpl
```diff
@@ -20,11 +20,6 @@
         "{{ .Resource.LabelType }}": {
             "{{ .Resource.LabelName }}": {
                 {{ .Resource.Body | printf "%s" }}
-                {{ if .Lifecycle.PreventDestroy -}}
```
turkenh, line 23: Are these relevant to this PR? Or just a refactor?

turkenh (approved):
LGTM! Thanks @muvaf !

pkg/terraform/resource/configuration.go
muvaf: Named the type as `ExternalNameConfiguration` but kept the field name in `Configuration` as `ExternalName`

pkg/tfcli/templates/main.tf.json.tpl
muvaf: Here we actually do some string manipulation on the printed form of the JSON, see the `,` before `prevent_destroy`. We'd have to unmarshal/marshal from `[]byte` to avoid doing that but now that we get to work with `map[string]interface{}`, we don't need any of this anymore. The additional property added directly to the map which is printed as is.

muvaf:
We synced with @turkenh about the `SelfVarPath` and how we could avoid it in the context of reference configurations. My main concern was that if we use a `map[string]resource.Configuration` in `apis` package,  then it'd mean that the `resource.Configuration` object would be limited to what can be imported in `apis` package which is quite limited. But @turkenh brought up the idea that we can have another struct that includes `resource.Configuration` so that its scope would be limited to only API type operations only, which would eliminate my concern.

The map will be similar to scheme accumulation in `apis` package, we decided to do it as part of https://github.com/crossplane-contrib/terrajet/issues/56

PR #62, opened 2021-09-13
Support configuring Terraform provider source & version via env. variables

ulucinar:
### Description of your changes

Please see: https://github.com/crossplane-contrib/terrajet/issues/51#issuecomment-918065123

I have:

- [X] Read and followed Crossplane's [contribution process].
- [X] Run `make reviewable` to ensure this PR is ready for review.
- [ ] Added `backport release-x.y` labels to auto-backport this PR if necessary.

### How has this code been tested

[contribution process]: https://git.io/fj2m9

PR #68, opened 2021-09-21
Read provider source and version with ProviderSetup function

### LINKED ISSUE #51: Capture Terraform Provider Source and Version during build time and make available in tfcli runtime

tfcli runtime needs the Terraform provider [source](https://github.com/crossplane-contrib/terrajet/blob/76069f71e80a5d234366fac60a21e5c40036ffc6/pkg/tfcli/options.go#L94) and [version](https://github.com/crossplane-contrib/terrajet/blob/76069f71e80a5d234366fac60a21e5c40036ffc6/pkg/tfcli/options.go#L102) at runtime for correctly generating the corresponding provider block in per-resource Terraform configuration. We need to capture this information during container image build time and make it available to the tfcli runtime.

The source of truth for the provider source and version constraint could be two env. variables, as both of pieces of information are also needed when preparing the local Terraform image mirror. Then as discussed previously, we could read those env. variables injected  into container's runtime in the `main` function, and have their values flow into tfcli runtime via the following path: `main -> (generated) controller.Setup -> <(generated) resource>.Setup -> terraform.SetupController` and stored in `terraform.connector` to be used by tfcli.

turkenh:

This PR:

- Reverts recently introduced [workaround](https://github.com/crossplane-contrib/terrajet/pull/62) to pass provider requirements via env vars.
- Extends (and renames as ProviderSetupFn) existing ProviderConfigFn to also pass provider source and version

Fixes #51 

I have:

Tested with https://github.com/crossplane-contrib/provider-tf-aws/pull/20

### DIFF
--- a/pkg/terraform/controller.go
+++ b/pkg/terraform/controller.go
@@ -40,25 +40,25 @@ const (
 	errUnexpectedObject = "the managed resource is not an Terraformed resource"
 )
 
-// ProviderSetupFn is a function that returns provider specific configuration
-// like provider credentials used to connect to cloud APIs.
-type ProviderSetupFn func(ctx context.Context, client client.Client, mg xpresource.Managed) (tfcli.ProviderRequirement, tfcli.ProviderConfiguration, error)
+// SetupFn is a function that returns Terraform setup which contains
+// provider requirement, configuration and Terraform version.
+type SetupFn func(ctx context.Context, client client.Client, mg xpresource.Managed) (tfcli.TerraformSetup, error)
 
 // NewConnector returns a new Connector object.
-func NewConnector(kube client.Client, l logging.Logger, ps ProviderSetupFn) *Connector {
+func NewConnector(kube client.Client, l logging.Logger, sf SetupFn) *Connector {
 	return &Connector{
-		kube:          kube,
-		logger:        l,
-		providerSetup: ps,
+		kube:           kube,
+		logger:         l,
+		terraformSetup: sf,
 	}
 }
 
 // Connector initializes the external client with credentials and other configuration
 // parameters.
 type Connector struct {
-	kube          client.Client
-	providerSetup ProviderSetupFn
-	logger        logging.Logger
+	kube           client.Client
+	logger         logging.Logger
+	terraformSetup SetupFn
 }
 
 // Connect makes sure the underlying client is ready to issue requests to the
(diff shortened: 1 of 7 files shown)


pkg/tfcli/options.go
```diff
@@ -220,16 +202,24 @@ func (r Resource) GetAddress() string {
...
 }
 
+// ProviderRequirement holds values for the Terraform HCL provider requirements
+type ProviderRequirement struct {
+	Source  string
+	Version string
+}
+
+// ProviderConfiguration holds the provider configuration body
+type ProviderConfiguration []byte
```
muvaf, line 212:
```suggestion
type ProviderConfiguration struct {
	Source  string
	Version string
	Credentials map[string]interface{}
}
```

If we pass around this single object, we'd be leaking less information about where it ends up. From Crossplane provider developer perspective, they don't need to know whether it's called requirement or configuration. We can hide this until go template section.

While I was working with terraform state stuff, I find it more handy to pass around `map[string]interface{}` instead of `[]data` so that we can make changes more easily down the line.

muvaf (commented):
I wonder if we can simplify the flow here along with https://github.com/crossplane-contrib/provider-tf-aws/pull/20#pullrequestreview-761088494 . It feels like we could achieve the same result if we hide more of the Terraform language until we print the HCL.

pkg/tfcli/options.go
turkenh: I am not sure. I understand the intent of hiding our implementation details but I believe by following similar naming it would be easier for developers to clearly see which struct expects what:

https://www.terraform.io/docs/language/providers/requirements.html
https://www.terraform.io/docs/language/providers/configuration.html

Also `Credentials` is not correct here, because it could contains other information in [provider configurations](https://www.terraform.io/docs/language/providers/configuration.html) which are usually not credentials (e.g. region)

turkenh: But as a simplification (for passing around a single object) I would be ok with the following:

```
type ProviderConfiguration map[string]interface{}

type ProviderSetup struct {
	ProviderRequirement
        ProviderConfiguration
}
```

turkenh: @muvaf wdyt?

muvaf: Looks good to me.

muvaf: @turkenh we need[ `TERRAFORM_VERSION`](https://github.com/crossplane-contrib/terrajet/pull/72/files#diff-68d8aa09b23aabebdc356ea5a16ebfa926b18b454366ac4c50f43ab8e8429178R82) as well during the reconstruction of the tfstate. `ProviderSetup` you suggested could be a good candidate to have this as field.

turkenh: Ok, I'll add it.

turkenh: Hmm, TerraformVersion didn't look great under `ProviderSetup` 🤔 

```
type ProviderSetup struct {
	TerraformVersion string
	Requirement      ProviderRequirement
	Configuration    ProviderConfiguration
}
``` 

Wondering if the following would be better: 

```
type TerraformSetup struct {
	Version          string
	Requirement      ProviderRequirement
	Configuration    ProviderConfiguration
}
```

turkenh: Going with the latter.

turkenh:
@muvaf just pushed the latest state after testing. Should be ready for merge.

pkg/terraform/controller.go
```diff
@@ -40,25 +40,25 @@ const (
...
 }
 
 // Connector initializes the external client with credentials and other configuration
 // parameters.
 type Connector struct {
-	kube           client.Client
-	providerConfig ProviderConfigFn
-	logger         logging.Logger
+	kube          client.Client
+	providerSetup SetupFn
```
muvaf, line 60:
```suggestion
	terraformSetupFn SetupFn
```

muvaf (approved):
LGTM! Thanks @turkenh !

Issue #61, opened 2021-09-13
Pipeline abstraction


muvaf:
### What problem are you facing?

In order to implement a code generator pipeline using Terrajet, authors need to wire up [bunch of generators](https://github.com/crossplane-contrib/provider-tf-aws/blob/c81d8a16ee8479bf1e3757c012c6d5f921aa3e2a/cmd/generator/main.go#L97). While this allows flexibility, it's not really necessary for some cases.

### How could Terrajet help solve your problem?

We can have a `Pipeline` struct that can accept options for customization while keeping the current generators for advanced cases. However, we need to solidify the initial set of providers to have a close-to-complete idea of what kinds of options we'd like to expose. After a certain threshold, it may not be worth to have the abstraction.

PR #69, opened 2021-09-23
Generate reference fields & custom configuration API

### LINKED ISSUE #56: Generate reference and selector fields if reference marker exists

turkenh:

This PR introduces a custom configuration that could be used to configure LateInitializationConfiguration and ReferencingConfiguration while generating the CRD. This PR also handles generating reference fields as well as passing required markers down to crossplane-tools to [generate](https://github.com/crossplane/crossplane-tools/pull/35) reference resolver functions.

Fixes #56

I have:

Tested with https://github.com/crossplane-contrib/provider-tf-aws/pull/22


pkg/comments/comment.go
```diff
@@ -4,52 +4,79 @@ import (
...
+	"github.com/crossplane-contrib/terrajet/pkg/terraform/resource"
 )
 
 // Option is a comment option
 type Option func(*Comment)
 
-// WithReferenceTo returns a comment options with reference to input type
-func WithReferenceTo(t string) Option {
+// WithReferenceConfig returns a comment options with the given reference config
+func WithReferenceConfig(cfg resource.FieldReferenceConfiguration) Option {
```
muvaf, line 14: Do you think we'd need other options? I feel like in general, people would just give the whole struct for reference, setting fields in one call.

pkg/pipeline/templates/terraformed.go.tmpl
```diff
@@ -38,7 +38,7 @@ func (tr *{{ .CRD.Kind }}) GetParameters() (map[string]interface{}, error) {
 	}
 	base := map[string]interface{}{}
 	{{- if .CRD.ConfigureWithFn }}
-	{{ .CRD.ConfigureWithFn }}.Configure(base, meta.GetExternalName(tr))
+	{{ .CRD.ConfigureWithFn }}(base, meta.GetExternalName(tr))
```
muvaf, line 41:
```suggestion
	// ConfigStoreBuilder is the one defined in CRD package.
	if cfg, ok := ConfigStoreBuilder[tr.GetTerraformResourceType()]; ok {
		// call the function in external name configuration
	}
```

I think the point of having a map was to be able to remove the function path as input. Would we be able to check the map and call the function?

pkg/types/reference.go
```diff
@@ -0,0 +1,100 @@
...
+		panic(errors.Wrap(err, "cannot build new comment for reference fields"))
+	}
+	req := false
+	commentOptional.KubebuilderOptions.Required = &req
+}
+
+// PathForType returns go package path for the input type. This is a helper
+// function to be used whenever this information is needed, like configuring to
+// reference to a type.
+func PathForType(i interface{}) string {
```
muvaf, line 98:
```suggestion
func TypePath(i interface{}) string {
```

nitpick

pkg/types/reference.go
```diff
@@ -0,0 +1,100 @@
...
+)
+
+var typeXPRef types.Type
+var typeXPSelector types.Type
+var commentOptional *comments.Comment
+
+func (g *Builder) getReferenceFields(t *types.TypeName, f *types.Var, r resource.FieldReferenceConfiguration) ([]*types.Var, []string) {
+	if r.ReferenceToType == "" {
+		return nil, nil
+	}
```
muvaf, line 33: We wouldn't need this check if this function is called only if there is a reference config in the map.

pkg/types/builder.go
```diff
@@ -21,6 +21,9 @@ import (
 	"go/token"
 	"go/types"
 	"sort"
+	"strings"
+
+	"github.com/crossplane-contrib/terrajet/pkg/terraform/resource"
```
muvaf, line 26: goimports

pkg/terraform/resource/store.go
```diff
@@ -0,0 +1,43 @@
+package resource
+
+// ConfigStore stores a global configuration
+type ConfigStore struct {
+	config map[string]*Configuration
+}
```
muvaf, line 6:
```suggestion
type ConfigStore map[string]Configuration
```

If we do that, we can use the pattern of `if cfg, ok := configMap[resourceName]; ok` easily. `nil` check is also possible but that could be simpler to work with.

pkg/types/reference.go
```diff
@@ -0,0 +1,100 @@
...
+	PackagePathXPCommonAPIs = "github.com/crossplane/crossplane-runtime/apis/common/v1"
+
+	loadMode = packages.NeedName | packages.NeedImports | packages.NeedTypes
+)
+
+var typeXPRef types.Type
+var typeXPSelector types.Type
+var commentOptional *comments.Comment
+
+func (g *Builder) getReferenceFields(t *types.TypeName, f *types.Var, r resource.FieldReferenceConfiguration) ([]*types.Var, []string) {
```
muvaf, line 30: I think we can start writing unit tests for utilities we don't mean to revisit. Could you add some for this function?

pkg/types/reference.go
```diff
@@ -0,0 +1,100 @@
...
+	// PackagePathXPCommonAPIs is the go path for the Crossplane Runtime package
+	// with common APIs
+	PackagePathXPCommonAPIs = "github.com/crossplane/crossplane-runtime/apis/common/v1"
+
+	loadMode = packages.NeedName | packages.NeedImports | packages.NeedTypes
+)
+
+var typeXPRef types.Type
+var typeXPSelector types.Type
+var commentOptional *comments.Comment
```
muvaf, line 28: nitpick
```suggestion
var TypeReferenceField types.Type
var TypeSelectorField types.Type
var CommentOptional *comments.Comment
```

muvaf (commented):
Great work! There are many comments but most of them are just nitpicks.

pkg/types/reference.go
muvaf: nitpick

```suggestion
func (g *Builder) GenerateReferenceFields(t *types.TypeName, f *types.Var, r resource.FieldReferenceConfiguration) ([]*types.Var, []string) {
```

pkg/terraform/resource/configuration.go
```diff
@@ -81,6 +70,20 @@ type ExternalNameConfiguration struct {
...
+// referenced.
+type ReferenceConfiguration map[string]FieldReferenceConfiguration
+
+// FieldReferenceConfiguration represents the Crossplane options used to generate
+// reference resolvers
+type FieldReferenceConfiguration struct {
+	ReferenceToType            string
+	ReferenceExtractor         string
+	ReferenceFieldName         string
+	ReferenceSelectorFieldName string
```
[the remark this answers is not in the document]
muvaf: And comments would be great, especially noting the optional fields.

pkg/types/builder.go
```diff
@@ -114,8 +121,15 @@ func (g *Builder) buildResource(res *schema.Resource, names ...string) (*types.N
 			req := !sch.Optional
 			comment.Required = &req
 			paramFields = append(paramFields, field)
-			g.comments.AddFieldComment(paramName, field.Name(), comment.Build())
 		}
+
+		fp := strings.Join(append(names[1:], field.Name()), ".")
```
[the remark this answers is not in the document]
muvaf: Actually, I think we should use `<Type Name>.<Field Name>` as the format of the key here because the same struct could be used elsewhere.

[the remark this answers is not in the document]
turkenh: I wouldn't prefer to include <Type Name> because it would require setting additional (already known) information in the custom configuration API.

I mean:

```
var rdsClusterCustomConfig = &resource.Configuration{
	ExternalName: resource.ExternalNameConfiguration{
		// External name config for RDS Cluster
	},
	References: map[string]resource.ReferenceConfiguration{
	--	"S3Import.BucketName": {
	--	"RDSCluster.S3Import.BucketName": {
			Type: types.PathForType(s3v1alpha1.S3Bucket{}),
		},
	},
}
```

pkg/types/reference.go
```diff
@@ -0,0 +1,100 @@
...
+}
+
+func init() {
+	pkgs, err := packages.Load(&packages.Config{Mode: loadMode}, PackagePathXPCommonAPIs)
+	if err != nil {
+		panic(errors.Wrap(err, "cannot load crossplane-runtime package to get reference types"))
+	}
+	if len(pkgs) != 1 && pkgs[0].Name != "v1" {
+		panic(errors.Wrapf(err, "unexpected package name %s", pkgs[0].Name))
+	}
```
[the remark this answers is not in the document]
turkenh: This is indeed not the case, since we use an explicit package path as pattern: "github.com/crossplane/crossplane-runtime/apis/common/v1". 

Just tested the following:

- setting pattern as  "github.com/crossplane/crossplane-runtime/pkg/resource", it returned only 1 package which is `resource` package. 
- setting pattern as "github.com/crossplane/crossplane-runtime/pkg/resource/...", it returned 6 packages, `resource`, `fake`, `unstructured` ...

So, we only want one package to be loaded and verified that one package is loaded correctly here.

pkg/types/reference.go
turkenh: Do you want to make this public, even only used in the same package? Why?

pkg/types/reference.go
turkenh: Do you want to make these public, even only used in the same package? Why?

pkg/pipeline/templates/terraformed.go.tmpl
turkenh: The problem with that is we would need to create an instance of ConfigStore and pass it to each controller at provider runtime as well. Currently, that map is initialized and used only during generation. 

It should not be a big deal and we can do it but I gave up doing that when I realize we would anyway need to pass string representation of reference types and extractor function, so, (we also synced on this) it is better to be consistent with external name function as well and take string representation of it as input which would mean no dependency on this ConfigStore object at runtime.

pkg/terraform/resource/store.go
turkenh: I believe we will end up with more configurations here that are not resource-specific (e.g. provider-specific). So, I would prefer something like the following to leave some room for future configurations.

```
type ConfigStore struct {
	resourceConfig map[string]Configuration
}
```

pkg/terraform/resource/configuration.go
```diff
@@ -97,8 +100,18 @@ type Configuration struct {
...
 	TerraformIDFieldName string
 }
+
+// OverrideConfiguration merges existing configuration with the input
+// configuration by overriding the existing one.
+func (c *Configuration) OverrideConfiguration(o *Configuration) error {
+	if o == nil {
+		return nil
+	}
+	return mergo.Merge(c, *o, mergo.WithOverride)
```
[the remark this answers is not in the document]
turkenh: I would prefer to keep it as it is not to worry about this while extending the configuration.

pkg/types/builder.go
```diff
@@ -114,8 +121,15 @@ func (g *Builder) buildResource(res *schema.Resource, names ...string) (*types.N
...
+		comment.FieldReferenceConfiguration = reference[fp]
```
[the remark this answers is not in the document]
turkenh: I don't see much of a value given that the field is already public.

pkg/terraform/resource/store.go
turkenh: I did some refactoring and moved configuration to a dedicated package (config), which is also a preparation for #61 where developers could provide configurations like which groups to generate, which resources to skip as part of this.

pkg/terraform/resource/store.go
```diff
@@ -0,0 +1,43 @@
...
+	cs.config[resource] = cfg
+}
+
+// GetConfigForResource gets the configuration for a given resource
+func (cs *ConfigStore) GetConfigForResource(resource string) *Configuration {
+	return cs.config[resource]
+}
+
+// ConfigStoreBuilder collects functions that add things to ConfigStore.
+type ConfigStoreBuilder []func(*ConfigStore) error
```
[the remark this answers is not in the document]
turkenh: Just pushed a new commit refactoring to do it in a simpler way as you suggested.

pkg/pipeline/templates/terraformed.go.tmpl
turkenh: With the latest refactoring, ignore the creating ConfigStore instance part. The only concern is consistency.

turkenh:
@muvaf I've resolved/responded to all your comments.

pkg/pipeline/templates/terraformed.go.tmpl
muvaf: Sounds good. It seems like the each configuration would work best with different format but that could be a pain to work with developers.

pkg/types/reference.go
muvaf: I guess it's my personal tendency, there isn't a solid reason to make them public.

pkg/types/builder.go
```diff
@@ -89,11 +96,19 @@ func (g *Builder) buildResource(res *schema.Resource, names ...string) (*types.N
...
-		fieldType, err := g.buildSchema(sch, append(names, fieldName.Camel))
+		fieldType, err := g.buildSchema(sch, refs, append(names, fieldName.Camel))
 		if err != nil {
 			return nil, nil, errors.Wrapf(err, "cannot infer type from schema of field %s", fieldName.Snake)
 		}
+
+		// Build the field path based on the fact that names is a slice of
+		// strings which contains the name of the parent fields sequentially.
+		// We skip the first element since it is the main CRD type.
+		fp := strings.Join(append(names[1:], fieldName.Camel), ".")
```
muvaf, line 107: We discussed online about whether having users give the field path or path in the `<type>.<field>` format which is what we need at minimum. It seems like it would work in both cases. For consistency of inputs, we're going with this implementation.

muvaf (approved):
The interface for configuration turned out to be challenging to get right with configurations spanning multiple mechanisms. @turkenh Thank you for working through all of them! I'm excited to see all come together under pipeline abstraction.

PR #71, opened 2021-09-23
tfcli: use json output of terraform plan to parse changes

muvaf:

The log of `terraform plan` gives enough information about what's planned that we don't need to rely on the returned code to assess whether we are up-to-date or not. I have been confused about what's happening here when I was working on state reproduce issue.

I have:

Manually with provider-tf-aws.

### DIFF
--- a/pkg/tfcli/refresh.go
+++ b/pkg/tfcli/refresh.go
@@ -20,23 +20,24 @@ import (
 	"context"
 	"fmt"
 	"path/filepath"
-	"regexp"
-	"strconv"
 	"strings"
 
 	"github.com/pkg/errors"
 	"go.uber.org/multierr"
 
+	"github.com/crossplane-contrib/terrajet/pkg/json"
 	"github.com/crossplane-contrib/terrajet/pkg/tfcli/model"
 )
 
 const (
 	// error messages
-	errRefresh        = "failed to refresh the Terraform state"
-	errNoID           = errRefresh + ": cannot deduce resource ID"
-	errImport         = "failed to import resource"
-	fmtErrRefreshExit = errRefresh + ": Terraform pipeline exited with code: %d"
-	fmtErrImport      = errImport + ": %s"
+	errRefresh      = "failed to refresh the Terraform state"
+	errNoID         = errRefresh + ": cannot deduce resource ID"
+	errImport       = "failed to import resource"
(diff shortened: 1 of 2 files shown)


pkg/tfcli/refresh.go
```diff
@@ -103,52 +104,64 @@ func (c *Client) importResource(ctx context.Context, id string) (model.RefreshRe
...
-		return result, err
+		return model.RefreshResult{}, err
+	}
+	stderr, err := c.pInfo.StderrAsString()
+	if err != nil {
+		return model.RefreshResult{}, err
 	}
+	// Code 2 means that we need to make a change and that's a valid case for
+	// us, so we don't treat it as an error case.
 	if code != 0 && code != 2 {
```
muvaf, line 128: When it returns code 2 or 0, the regex wouldn't work

pkg/tfcli/refresh.go
```diff
@@ -103,52 +104,64 @@ func (c *Client) importResource(ctx context.Context, id string) (model.RefreshRe
...
-		return result, errors.Errorf(fmtErrRefreshExit, code)
+		return model.RefreshResult{}, errors.Errorf(errFmtObserve, stderr)
```
turkenh, line 129: Should we still return exit code as part of the error?

turkenh (approved):
I think I reviewed this as part of #72 but adding the same comment here. 
Other than that it looks good to me. 
But I would suggest waiting for @ulucinar to give a thumbs up here.

pkg/tfcli/refresh.go
muvaf: done!

muvaf:
@ulucinar could you take a look?

ulucinar (approved):
LGTM. I liked using the `-json` command-line option that produces log messages that are suitable for programmatic access, instead of the non-json format. I was not aware of that option.

One idea might be to rely on the exit code of the `terraform apply` command to deduce whether the resource is up-to-date (as it was the case before this change) because exit codes make better interfaces between processes than parsing log output. However, unfortunately we still need to parse terraform CLI output to learn whether the external resource is missing or needs an update, so no objections for inferring up-to-date resources also from the command output, especially when parsing JSON. We may assume JSON schema will be more stable.

If at some point, terraform implements better semantics for their exit codes, then we can consider getting rid of log parsing.

Issue #12, opened 2021-08-13
Alternative ways to store terraform state


turkenh:
In the initial implementation, we are planning to store terraform states as annotations for the managed resource.

ETCD (i.e. annotation) will be authoritative in terms of the state and we avoid making any changes/edits on it except stripping out sensitive attributes. This is the most conservative approach in terms of using terraform just like as an end-user.

However, we want to investigate whether we can safely (re)build the state information from the spec, status, and kind information of the resource and completely remove the annotation.

To better describe the problem, see the following `tfstate` example:

```
{
  "version": 4,
  "terraform_version": "1.0.4",
  "serial": 5,
  "lineage": "f36b8453-5686-97c3-af3e-6a5c655c5fbf",
  "outputs": {},
  "resources": [
    {
      "mode": "managed",
      "type": "aws_vpc",
      "name": "instance_vpc_hasan",
      "provider": "provider[\"registry.terraform.io/hashicorp/aws\"]",
      "instances": [
        {
          "schema_version": 1,
          "attributes": {
            "arn": "arn:aws:ec2:us-west-2:609897127049:vpc/vpc-0b396ddd9469f29e5",
            "assign_generated_ipv6_cidr_block": false,
            "cidr_block": "10.0.0.0/16",
            "default_network_acl_id": "acl-08e8f435b74af56e8",
            "default_route_table_id": "rtb-0eb8f2b9933d31175",
            "default_security_group_id": "sg-034d7899a57871f29",
            "dhcp_options_id": "dopt-0c388774",
            "enable_classiclink": false,
            "enable_classiclink_dns_support": false,
            "enable_dns_hostnames": false,
            "enable_dns_support": true,
            "id": "vpc-0b396ddd9469f29e5",
            "instance_tenancy": "default",
            "ipv6_association_id": "",
            "ipv6_cidr_block": "",
            "main_route_table_id": "rtb-0eb8f2b9933d31175",
            "owner_id": "609897127049",
            "tags": {
              "Name": "ExampleVPCInstanceByHasan"
            },
            "tags_all": {
              "Name": "ExampleVPCInstanceByHasan"
            }
          },
          "sensitive_attributes": [],
          "private": "eyJzY2hlbWFfdmVyc2lvbiI6IjEifQ=="
        }
      ]
    }
  ]
}
```

- `resources[0].instances[0].attributes` field, is a combination of
  - `spec.forProvider`
  - `status.atProvider`
  - `external_name` annotation (`id` here)
[...]

Issue #36, opened 2021-08-26
Discuss filesystem locks and the need for repeated calls to get the operation status


muvaf:
### What problem are you facing?

During the initial implementation, in order to move fast and see our limits we deferred some of the discussion about the implementation and how we should do the bookkeeping was one of them. Currently, we manage the state and bookkeeping using [filesystem](https://github.com/crossplane-contrib/terrajet/blob/9b4eaff67cf52c1d97725d1a8f7661fef840d77d/pkg/tfcli/client.go#L142). However, we also interact with Terraform's lock and state files, which makes the code complex to understand for readers and we don't really need to keep that information out of memory since the crash of the process means a restart of the `Pod` anyway. Additionally, there are some opportunities to merge some of the functionality in the code to get a simpler flow that is easier to read and change.

### How could Terrajet help solve your problem?

I suggest that we use a global map that does the bookkeeping of the running CLI executions. Roughly, I think we need the following improvements in the initial implementation. Please feel free to add more things that we had deferred to after PoC phase. cc @ulucinar @turkenh 

* A global async-safe map that holds information about existing Terraform workspaces, like `map[string]Workspace` whose key is `metadata.uuid`.
  * Operation, which is invoked with a method on `Workspace`, should update the map before it starts and after it completes, similar to today.
* `Describe` method on that `Workspace` object that will tell you the current status, i.e. `creating`, `destroying` or `no-op`.
  * For example, we report that the resource does not exist during creation because we need `apply` pipeline to be run repeatedly to check the status. But no other provider does that except some of Azure APIs, which is not user friendly at all in a controller, so we run into possibility of violating generic reconciler assumptions.
  * Today, we make controller call `Create` repeatedly because we need to make the same `Apply` call to finalize it but since https://github.com/crossplane/crossplane-runtime/pull/283 , this won't be allowed because generic reconciler won't call `Create` again in the grace period.
* This map `map[string]Workspace` can be treated just like a cloud provider SDK client. We don't need to know Terraform CLI, pipeline/operation details or propagating status as error etc. We can use `map[string]interface{}` to pass down spec configuration and receive `StateV4` and information about whether it's `creating` or `destroying`.
  * We could possibly remove the conversion layer completely and have the `ExternalClient` implementation directly call a single interface. Because if you look at it today, it's merely a shim on top of conversion layer. Since we use similar names, it makes it really hard to follow the layers.
[...]

Issue #65, opened 2021-09-13
Sensitive data leaks at `status.atProvider` and the "tf.crossplane.io/state" annotation


ulucinar:
### What happened?

When testing provider-tf-azure's `KubernetesCluster` MR, we have observed that provisioned cluster's credentials, e.g. KubeConfig, are available in the Teffaform state as regular fields. An example is:
```json
{
  "version": 4,
  "terraform_version": "1.0.5",
  "serial": 3,
  "lineage": "365fcd38-09a3-e30a-cb02-df3117306af0",
  "outputs": {},
  "resources": [
    {
      "mode": "managed",
      "type": "azurerm_kubernetes_cluster",
      "name": "example",
      "provider": "provider[\"registry.terraform.io/hashicorp/azurerm\"]",
      "instances": [
        {
          "schema_version": 0,
          "attributes": {
...
            "kube_config": [
              {
                "client_certificate":
```

Because these attributes are resource dependent, we need to be able to treat them as credentials in a resource specific manner. It looks like in the Terraform provider resource schema, such fields are marked as `sensitive`, like [kube_config_raw](https://github.com/hashicorp/terraform-provider-azurerm/blob/d347dedf86f382a5a614710aa41ab6445b8519f9/internal/services/containers/kubernetes_cluster_resource.go#L869), or [kube_config.password](https://github.com/hashicorp/terraform-provider-azurerm/blob/d347dedf86f382a5a614710aa41ab6445b8519f9/internal/services/containers/kubernetes_cluster_resource.go#L845).

Those fields that are marked as `sensitive` are also good candidates for being part of the [connection secrets](https://github.com/crossplane-contrib/terrajet/issues/35). 

### How can we reproduce it?

Provisioning a `KubernetesCluster` using https://github.com/crossplane-contrib/provider-tf-azure/blob/258830e0bc0627f2749bcd7e285e3c831849a04a/examples/kubernetes/kubernetescluster.yaml reveals the issue.

muvaf:
I think https://github.com/crossplane-contrib/terrajet/issues/35 would cover this. Closing. Please reopen if that's not the case.

Issue #35, opened 2021-08-26
Figure out how to handle sensitive information for managed resources


turkenh:
### What problem are you facing?

Crossplane provides a way to connect to cloud resources (if applicable) by writing the required details to connect into a Kubernetes secret. 
These details should include all fields that would need to be required to connect including sensitive ones like `username` and `password` and not necessarily sensitive ones like `port` information. 

### How could Terrajet help solve your problem?

We would need to define/implement a way to generate these connection details with terrajet. 
I am assuming this information could be composed using the `attributes` and `sensitive_attributes` fields in terraform state.

muvaf:
I think there are two issues regarding sensitive information:
1. Output returned from the providers that is stored in tfstate.
2. Input from user as a configuration parameter.

The tricky part is that there are cases where without implementing (2), we can't really remove the field. For example, RDS instance has a field to take password. We can copy the content of `password` field from tfstate into our connection detail but we can't remove the field from schema because it's used as input to the creation call. 

We can possibly either expand the scope of this issue to also generate and process `*v1.SecretKeyReference` fields hence remove the sensitive fields completely, or keep the scope as is and don't remove the fields, only make sure that their content in tfstate makes it to the connection detail.

turkenh:
Good point. Originally, I was thinking the scope of this issue is more around the first one. 
But the second could be solved by following a similar solution as we used to mark and generate reference fields, e.g. developer marks the field as sensitive and terrajet adds a secretref field if it is input together with providing function to read it. I don't think we would need a special code generation like reference resolvers though since this case would be simpler.

if field is marked as sensitive & input (i.e. from spec) => generate a SecretKeyRef field and read from the referenced secret in GetParameters. 
if field is marked as sensitive & output (i.e. to status) => store in connection details secret in SetObservation (may be also store input secret to connection details).

Considering the possible flow above, I think it makes sense to handle both cases with this ticket.

muvaf:
> generate a SecretKeyRef field and read from the referenced secret in GetParameters.
> store in connection details secret in SetObservation (may be also store input secret to connection details).

We'd have to add `kube client.Client` argument to `GetParameters` function and return a `map[string][]byte` from `SetObservation`, right?

I think it starts to feel like we're overloading the functions that seem very basic on the outside, i.e. classic getters and setters, with some logic that are normally handled elsewhere in manually written controllers.

I wonder if we can extend `Terraformed` interface call those functions in the controller itself. For example:
```go
[...]

PR #72, opened 2021-09-23
Reproduce state in every reconciliation using the managed resource fields

muvaf:

With this change, we are able to reproduce tf state in every reconcile. Mechanism is roughly as the following:
* Dump parameters, observation and ID into state.
  * We dump everything we have because `terraform import` does not cover all information, like the non-sensitive values that are available only during creation.
* Run `terraform refresh` to get the current state of the world, which will update only the things that don't match.
  * At this point, we have the current state as tfstate file and desired state in our CR parameters.

We had to store [`private` attribute](https://serverfault.com/questions/981661/what-is-the-meaning-of-private-attribute-in-tfstate-for-a-dynamodb-instance) of the state object (usually a simple json with schema version info) in annotations because it's arbitrary metadata information that provider uses and cannot be regenerated since it's neither parameter nor observation.

While I'm working on this issue, I tried not to touch directly unrelated things but filled out https://github.com/crossplane-contrib/terrajet/issues/36 for those things and opened small other PRs for smaller refactorings.

Fixes https://github.com/crossplane-contrib/terrajet/issues/12

I have:

Creation and stable state:
* Create the resource.
* Delete workspace folder
* Check whether it's recreated and everything is working.

Update:
* Create the resource.
* Delete workspace folder.
* Update CR instance.
* Check whether workspace is reproduced and update is issued.


turkenh:
@muvaf I didn't make a review yet, but are we sure that this PR is fixing #65? My understanding from the PR description is, it would partially solve it by removing from annotations, but not `status.atProvider` (which I would expect to be fixed with #35) ?

muvaf:
@turkenh it doesn't, thanks for catching it!

pkg/tfcli/client.go
```diff
@@ -202,23 +191,46 @@ func (c *Client) syncPipeline(ctx context.Context, ignoreExitErr bool, command s
...
+			Type: c.resource.LabelType,
+			Name: c.resource.LabelName,
+			// TODO(muvaf): we should get the full URL from Dockerfile since
+			// providers don't have to be hosted in registry.terraform.io
+			ProviderConfig: fmt.Sprintf(`provider["registry.terraform.io/%s"]`, c.provider.Source),
+			Instances: []json.InstanceObjectStateV4{
+				{
+					SchemaVersion: 0,
+					PrivateRaw:    []byte(c.resource.PrivateRaw),
+					AttributesRaw: attr,
```
turkenh, line 223: We are not building `AttributeSensitivePaths` (`sensitive_attributes`) as part of the state? Are we confident that no resource would require some information from there in its state?

pkg/tfcli/options.go
```diff
@@ -32,58 +32,40 @@ import (
...
+
 	envTFProviderVersion = "XP_TERRAFORM_PROVIDER_VERSION"
 	envTFProviderSource  = "XP_TERRAFORM_PROVIDER_SOURCE"
 
 	fmtResourceAddress = "%s.%s"
+
+	// AnnotationKeyPrivateRawAttribute is the key that points to private attribute
+	// of the Terraform State. It's non-sensitive and used by provider to store
+	// arbitrary metadata, usually details about schema version.
+	AnnotationKeyPrivateRawAttribute = "terrajet.crossplane.io/private"
```
turkenh, line 44: I am afraid users might get confused by this annotation, thinking that we are storing something private (sensitive) as an annotation.
Could we find another alternative? 
After reading [this answer](https://serverfault.com/a/981750), which also validates my concern above :), may be we can use something like: `terrajet.crossplane.io/provider-meta`

pkg/tfcli/options.go
```diff
@@ -192,19 +181,18 @@ func (c Client) validate() error {
...
-// https://www.terraform.io/docs/language/meta-arguments/lifecycle.html
-type Lifecycle struct {
-	PreventDestroy bool
+	// This is a provider-specific metadata that varies between TF providers.
+	PrivateRaw string
 }
 
-func (r Resource) validate() error {
+func (r *Resource) validate() error {
 	if r.LabelName == "" || r.LabelType == "" {
```
turkenh, line 196: Could we add more validations here? like UID cannot be empty.

pkg/tfcli/refresh.go
```diff
@@ -19,136 +19,98 @@ package tfcli
...
+	errFmtParsePlan = "failed to parse Terraform plan output: %s"
 )
 
 // Refresh updates local state of the Cloud resource in a synchronous manner
-func (c *Client) Refresh(ctx context.Context, id string) (model.RefreshResult, error) {
-	if id == "" && len(c.tfState) == 0 {
-		return model.RefreshResult{}, errors.New(errNoID)
-	}
-
+func (c *Client) Refresh(ctx context.Context, _ string) (model.RefreshResult, error) {
```
turkenh, line 38: Do we need the second parameter in function signature?

pkg/tfcli/refresh.go
```diff
@@ -19,136 +19,98 @@ package tfcli
...
-	// as a precaution, remove any existing state file
-	if err := c.fs.RemoveAll(filepath.Join(c.wsPath, fileState)); err != nil {
-		return result, errors.Wrap(err, errImport)
-	}
-	// now try to run the synchronous import pipeline
-	if err := c.syncPipeline(ctx, true, pathTerraform, "import", "-input=false", c.resource.GetAddress(), id); err != nil {
+	// Prepare the workspace for a refresh operation because the state that comes
+	// with import may not include all information we have. So, we first write
+	// everything into state file and then run refresh to update the ones that
+	// can be updated with the current state of the world. Import cannot be ru
```
turkenh, line 45: nit:
```suggestion
	// can be updated with the current state of the world. Import cannot be run
```

pkg/tfcli/refresh.go
```diff
@@ -19,136 +19,98 @@ package tfcli
...
+	// with existing state file.
```
turkenh, line 46: Did we lose the import functionality?

pkg/tfcli/refresh.go
```diff
@@ -19,136 +19,98 @@ package tfcli
...
 	}
+	stderr, err := c.pInfo.StderrAsString()
+	if err != nil {
+		return model.RefreshResult{}, err
+	}
+	// Code 2 means that we need to make a change and that's a valid case for
+	// us, so we don't treat it as an error case.
 	if code != 0 && code != 2 {
-		return result, errors.Errorf(fmtErrRefreshExit, code)
+		return model.RefreshResult{}, errors.Errorf(errFmtObserve, stderr)
```
turkenh, line 77: Should we still return exit code as part of the error?

turkenh (commented):
Thanks for working on this Muvaffak! I believe this would help us to get closer to behave like a native crossplane provider.

pkg/tfcli/refresh.go
turkenh: Hmm, actually we didn't, we import just via refreshing state with enough fields.

pkg/terraform/controller.go
```diff
@@ -175,7 +181,7 @@ func (e *external) Delete(ctx context.Context, mg xpresource.Managed) error {
 // the object.
 func (e *external) persistState(ctx context.Context, obj xpresource.Object) error {
```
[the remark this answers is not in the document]
muvaf: Due to problem described in https://github.com/crossplane-contrib/terrajet/issues/36#issuecomment-925130302 , we are not able to use the latest crossplane-runtime yet.

pkg/tfcli/client.go
muvaf: No, we are not sure. But I haven't seen a resource put any data there so I don't really know where we can derive it from, its format or what it's used for to understand whether it's another thing to store in annotation.

pkg/tfcli/options.go
muvaf: Good call!

pkg/tfcli/refresh.go
muvaf: Yes we still have the import functionality. Since we removed the annotation, we don't have a signal as clear as `len(c.tfState) == 0` to decide whether we need to call import or not. Checking whether observation or parameters are empty do not always mean that the resource is not created or not in creating state because it could take a few reconciles until the creation is completed and observation reflects the state.

muvaf: Extended the comment to describe why we use refresh all the time https://github.com/crossplane-contrib/terrajet/pull/72/commits/cdf118898f9afa211b773e86a6c5765ceaffc3a6

muvaf:
@turkenh addressed all your comments and rebased.

pkg/tfcli/client.go
muvaf: It seems like it's only metadata https://github.com/hashicorp/terraform/issues/25646#issuecomment-718052143 . If it's just what's described here, then it's not really something we need to care about.

pkg/process/fork.go
```diff
@@ -36,6 +37,13 @@ const (
 	errNotBuffered    = "stream not buffered"
 )
 
+var (
+	// ansiEscaper is used to remove color characters that Terraform CLI uses
+	// in its output because they make reading the string challenging in the contexts
+	// where coloring is not available, such as custom resource or events.
+	ansiEscaper = regexp.MustCompile("[\u001B\u009B][[\\]()#;?]*(?:(?:(?:[a-zA-Z\\d]*(?:;[a-zA-Z\\d]*)*)?\u0007)|(?:(?:\\d{1,4}(?:;\\d{0,4})*)?[\\dA-PRZcf-ntqry=><~]))")
```
muvaf, line 44: Actually, I realized `-no-color` flag does that as well, I'll update the PR with that and remove this.

turkenh (approved):
lgtm!

ulucinar:
Please note that this comment does not imply I'm against the changes proposed in this PR. But I have some concerns about the direction of the PR that I'd like to mention. They are not fully substantiated concerns at the moment, and solely from a Crossplane point of view, I see the merit in the proposed changes.

We should be more careful as we diverge from what Terraform folks suggest us to do:
https://www.terraform.io/docs/language/state/purpose.html

As long as we can reproduce tfstate intact, we should be on the safe side from this regard, aside from Terraform bugs. And the recommended way to do this is to use [remote state](https://www.terraform.io/docs/language/state/remote.html). In our case, the main motivation is not teamwork but because the local filesystem where tfstate is stored is ephemeral. 

As we diverge from what's suggested, we have to make more operational assumptions that will break sometime. And this means that we are adding additional risks on top of existing ones. There are some risks that may not be possible to evade with Terraform-based providers, such as those associated with schema generation. But there is also a second* class of risks that we take (and perhaps are willing to take) solely because we diverge from the common, supported, well-documented usage patterns of the underlying Terraform technologies. 

Please note that the dimensions mentioned in the above document, "Mapping to the Real World", "Metadata", "Performance", and "Syncing" need not to be an exhaustive list. This set can change with new Terraform versions and even for versions that our released software depends on, they are difficult to validate. However, not all of them are relevant in our Crossplane-based environment either. For instance, the dependency-recording argument mentioned in the Metadata section is currently not relevant because we have a single resource specified in our Terraform configurations. Or the fundamental problem of mapping a Terraform configuration resource to the corresponding external resource (in Crossplane terms) should anyhow be addressed by correctly storing the external name annotation.

But do we know/handle every type of metadata that can be stored in tfstate for the current Terraform versions we depend on? Or what happens when those types change with new Terraform versions? We adapt our software to those prospective changes but how do we detect such maintenance needs, how do we test them? 

Performance is also a quite controversial topic in this context. The above document mentions about treating tfstate as the source of truth for the current state of the external resources in performance-sensitive environments by making heavy use of `-refresh=false`. This is something I believe currently not discussed in Crossplane as every provider implementation I'm aware of observes its external resources via its `ExternalClient.Observe` implementation. However, this should not also mean that we will never have to consider such implementations/options someday. 

As mentioned above, we cannot easily guarantee that these discussions are an exhaustive current set of potential issues if we don't treat tfstate as opaque. Again, I consider ourselves still experimenting and playing with ideas, so this comment is not a blocker for this PR. But I have a hunch that treating tfstate as opaque and remotely storing and may be carrying a reference to it in the MR might be a less risky approach. 

But software is ours, let's go and explore it :)

*: We can also think of more risk classes in this regard. For instance, for late-initialization or for setting `status.atProvider` in Crossplane resource model, we need some transparency into tfstate. We are already tackling with those associated issues.

muvaf:
Thank you for calling out the risks @ulucinar , very much appreciated! I think some of those concerns have been there inherently once we chose Terraform, i.e. they can do something that will break us on purpose or not, and we have made a strategic decision to rely on that. Now, as you mention, there are local risks that we take in order to make Terraform work for us in a way that we can provide all XRM features, including Kubernetes Resource Model.

I think if we keep the state in annotation, we'd be violating two rules:
* Observation of the world should be stored under `status` - [KRM](https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api-conventions.md#spec-and-status)
* Properties should not be duplicated in spec and status - [XRM](https://github.com/crossplane/crossplane/blob/master/design/one-pager-managed-resource-api-design.md#high-fidelity)

As mentioned in Terrajet design doc, [one of our goals](https://github.com/crossplane/crossplane/blob/master/design/design-doc-terrajet.md#goals) is to make Terrajet providers XRM-compliant to the extent that's possible and reliable.

I believe the risk imposed by this specific change is that there are [certain fields](https://github.com/crossplane-contrib/terrajet/issues/27) in schema that we were not able to add to neither spec nor status. But the minimum we need for Terraform to work is the ID since Terraform guarantees that the state that can be produced by ID is useful enough for it. The reason that we put all the fields we have is to make sure tfstate includes the fields that are not covered by import. So, where we stand is somewhere between the state produced by `terraform import/refresh` and the full tfstate that is kept since the beginning, being much close to the latter.

IMO, the risk of corruption in this flow seems minor to me and is worth taking in order to be more XRM-compliant. If at some point Terraform decides to make a change we'd have to update provider and Terraform version to be impacted. To your point that's where the testing infra can help at that point. I believe something like [localstack](https://github.com/localstack/localstack) for example could be pretty helpful for full end-to-end integration tests that include a set of resources that are general enough to cover most of the resources in terms of their behavior.

PR #77, opened 2021-09-29
Handle sensitive fields

### LINKED ISSUE #35: Figure out how to handle sensitive information for managed resources

turkenh:

Fixes #35 

On top of PR #69 

TODO:
- [x] Extract field paths for fields marked as sensitive in resource schema
- [x] Get values matching field paths
- [x] Developers can add custom fields to be included in addition to whats is coming with schema
- [x] Generate GetConnectionDetails function in terraformed and verify sensitive fields landing in connection details secret 
- [ ] Adjust CRD schema for sensitive fields
  - [ ] If the sensitive field is a parameter, generate secret key ref in the schema
  - [ ] If the sensitive field is an observation, drop the field from the schema
- [ ] Generate GetSensitiveParameters function in terraformed

I have:

- [ ] Read and followed Crossplane's [contribution process].


pkg/pipeline/templates/terraformed.go.tmpl
```diff
@@ -60,3 +61,14 @@ func (tr *{{ .CRD.Kind }}) SetParameters(params map[string]interface{}) error {
 	}
 	return json.TFParser.Unmarshal(p, &tr.Spec.ForProvider)
 }
+
+// GetConnectionDetails of this {{ .CRD.Kind }}
+func (tr *{{ .CRD.Kind }}) GetConnectionDetails(data []byte) (map[string][]byte, error) {
```
turkenh, line 66: I remember we switched using `map[string]interface{}` in other terraformed functions but now I am questioning whether it was the right decision. At least, for some functions (e.g. `SetObservation`, `SetParameters` and this one here), we need serialized data and what we do for them is 

`serialized (tfstate) -> unmarshall (map[string]interface) -> marshall -> unmarshall (to spec and status)`
instead of 
`serialized (tfstate) -> unmarshall (to spec and status)` 

Other than concerns like performance and some possibility of data loss, I think this is unnecessary. I would suggest using whichever makes sense (`serialized` vs `map[string]interface`) instead of forcing `map[string]interface `everywhere. 
@muvaf wdyt?

turkenh: Ok, please ignore this. I ended up using map[string]interface{} as well since more convenient.

pkg/terraform/resource/sensitive.go
```diff
@@ -0,0 +1,200 @@
...
+			}
+		}
+	}
+
+	return nil
+}
+
+// GetSensitiveObservation will return sensitive information as terraform state
+// attributes by reading them from connection details.
+func GetSensitiveObservation(from runtime.Object, into map[string]interface{}, mapping map[string]string) error {
```
turkenh, line 114: I need to fix this, sensitive observation should come from conn details secret not status.atProvider.

turkenh: Fixed!

pkg/types/builder.go
```diff
@@ -96,27 +99,58 @@ func (g *Builder) buildResource(res *schema.Resource, refs config.References, na
 		if comment.TerrajetOptions.FieldJSONTag != nil {
 			jsonTag = *comment.TerrajetOptions.FieldJSONTag
 		}
-		fieldType, err := g.buildSchema(sch, refs, append(names, fieldName.Camel))
+
```
muvaf, line 102: During the initial implementation, I was thinking to have a generic interface like `FieldProcessor` that we can call in a for loop here to process each field in the schema (like `Reference`, `Sensitive`, `Generic` etc) so that each behavior logic would be contained in its own field processor but didn't pursue it. It may not even be possible since there are a few independent things each processor would do and it could encompass more than one field.

Not related to this PR at all, but I'm curious about your thoughts about whether it'd be possible to do it at some point given that you have more context about the nature of the flow here.

pkg/terraform/files.go
```diff
@@ -48,12 +49,12 @@ func WithFileSystem(fs afero.Fs) FileProducerOption {
...
 // NewFileProducer returns a new FileProducer.
-func NewFileProducer(dir string, tr resource.Terraformed, ts Setup, opts ...FileProducerOption) (*FileProducer, error) {
-	params, err := tr.GetParameters()
+func NewFileProducer(ctx context.Context, c resource.SecretClient, dir string, tr resource.Terraformed, ts Setup, opts ...FileProducerOption) (*FileProducer, error) {
+	params, err := tr.GetParameters(ctx, c)
 	if err != nil {
 		return nil, errors.Wrap(err, "cannot get parameters")
 	}
-	obs, err := tr.GetObservation()
+	obs, err := tr.GetObservation(ctx, c)
```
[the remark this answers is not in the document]
turkenh: Yeah, sounds good in terms of simplifying Terraformed interface. 
Done!

pkg/resource/sensitive.go
```diff
@@ -0,0 +1,191 @@
...
+
+// GetSensitiveParameters will collect sensitive information as terraform state
+// attributes by following secret references in the spec.
+func GetSensitiveParameters(ctx context.Context, client SecretClient, from runtime.Object, into map[string]interface{}, mapping map[string]string) error {
+	pv, err := fieldpath.PaveObject(from)
+	if err != nil {
+		return err
+	}
+	xpParams := map[string]interface{}{}
+	if err = pv.GetValueInto("spec.forProvider", &xpParams); err != nil {
```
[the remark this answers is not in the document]
turkenh: Not sure I get exactly what you mean with _more correct._
In the mapping of the sensitive field, we keep `tfPath`: `xpPath` but we do not append "spec.forProvider" to xpPath path (would be obsolete)
By getting what is under `spec.forProvider`, we kind of normalize the XP parameters and use it. We could also prepend "spec.forProvider" each time we get the fields but I don't see much of a reason to do so.

pkg/resource/sensitive.go
```diff
@@ -0,0 +1,191 @@
...
+			}
+		}
+	}
+
+	return nil
+}
+
+// GetSensitiveObservation will return sensitive information as terraform state
+// attributes by reading them from connection details.
+func GetSensitiveObservation(ctx context.Context, client SecretClient, from *v1.SecretReference, into map[string]interface{}) error {
```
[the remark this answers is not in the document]
turkenh: I gave this a try but didn't like the fact that I would need to refactor all unit tests to pass a managed (or terraformed) resource even the method actually needs very little information from it. 
I would prefer to pass as little information as possible until we really need too much information that we would need to really pass terraformed (or managed)

pkg/resource/sensitive.go
```diff
@@ -0,0 +1,191 @@
...
+	paveTF := fieldpath.Pave(into)
+	for k, v := range conn {
+		if err = paveTF.SetString(k, string(v)); err != nil {
+			return errors.Wrapf(err, "cannot set sensitive string in tf attributes for key %q", k)
+		}
+	}
+	return nil
+}
+
+func expandedTFPath(expandedXP string, mapping map[string]string) (string, error) {
```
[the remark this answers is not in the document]
turkenh: Thanks for the suggestion and for preparing the code snippet.

I'll try if I can simplify following an approach you suggested, however, one caveat here is, fieldpath.Paved does not print keys of maps inside `[ ]` if the key contains dots. See [this](https://github.com/crossplane-contrib/terrajet/pull/77/files#diff-bcaded8cd4a1dc0b60666c12460b81c9abb6379f5374a09e478c505d486073d5R130) as an example. So, initially, I attempted some regex replace kind of solution but I ended up parsing them using the same parsing not to miss something or build a complex method to handle edge cases like I mentioned.

pkg/types/builder.go
turkenh: Yeah, I think that's would improve the readability and maintenance of this path of the codebase. I was also not feeling good about adding new lines to a single flow. Not sure how much we would need to add though.
This could be tried out while writing unit tests here may be.

pkg/types/builder.go
```diff
@@ -96,27 +99,58 @@ func (g *Builder) buildResource(res *schema.Resource, refs config.References, na
...
+		if ref, ok := cfg.References[tfFieldPath]; ok {
 			comment.Reference = ref
 			sch.Optional = true
 		}
-		field := types.NewField(token.NoPos, g.Package, fieldName.Camel, fieldType, false)
+
+		fieldNameCamel := fieldName.Camel
+		if e, ix := containsAt(cfg.Sensitive.CustomFieldPaths, tfFieldPath); e || sch.Sensitive {
+			if e {
+				cfg.Sensitive.CustomFieldPaths = remove(cfg.Sensitive.CustomFieldPaths, ix)
```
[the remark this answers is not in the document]
turkenh: This was something I considered but I didn't like that users would need to provide a map in that case and the value part of the map (xp/json path) is even unknown before the code generation. We could expect them to set some placeholder (or empty?) values but that sounded hacky. 
But anyway, as we discussed today, I will need to revisit getting custom configuration since we would need the distinction of `include into connection details but not sensitive` type of fields.

turkenh:
Thanks for the review @muvaf! Resolved/responded to all your comments. 

The two open items I need to add to PR is:
- Always include `id` in connection details
- Ability to configure fields to get into connection but not mark as sensitive (i.e. no schema change)

pkg/types/builder.go
```diff
@@ -96,27 +99,58 @@ func (g *Builder) buildResource(res *schema.Resource, refs config.References, na
...
+			}
+
+			if isObservation(sch) {
+				cfg.Sensitive.AddFieldPath(tfFieldPath, xpFieldPath)
+				// Drop an observation field from schema if it is sensitive.
+				// Data will be stored in connection details secret
+				continue
+			}
+			sfx := "SecretRef"
```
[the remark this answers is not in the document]
turkenh: Yeah, but not sure `xyzSecretRef` would be better in such a case or are there any better alternative

pkg/pipeline/templates/terraformed.go.tmpl
```diff
@@ -23,13 +22,21 @@ func (tr *{{ .CRD.Kind }}) GetTerraformResourceIDField() string {
 	return "{{ .Terraform.IdentifierField }}"
 }
 
+// GetConnectionDetailsMapping for this {{ .CRD.Kind }}
+func (tr *{{ .CRD.Kind }}) GetConnectionDetailsMapping() map[string]string {
+  return map[string]string{ {{range $k, $v := .SensitiveFields}}"{{ $k }}": "{{ $v}}", {{end}} }
```
muvaf, line 27: nitpick: return `nil` if `.SensitiveFields` doesn't exist?

pkg/pipeline/templates/terraformed.go.tmpl
```diff
@@ -23,13 +22,21 @@ func (tr *{{ .CRD.Kind }}) GetTerraformResourceIDField() string {
...
 func (tr *{{ .CRD.Kind }}) GetObservation() (map[string]interface{}, error) {
 	o, err := json.TFParser.Marshal(tr.Status.AtProvider)
 	if err != nil {
 		return nil, err
 	}
 	base := map[string]interface{}{}
+	if err := json.TFParser.Unmarshal(o, &base); err != nil {
+	  return nil, err
+	}
 	return base, json.TFParser.Unmarshal(o, &base)
```
muvaf, line 40:
```suggestion
	return base, json.TFParser.Unmarshal(o, &base)
```

Seems like this is remnant of the earlier implementation.

pkg/pipeline/templates/terraformed.go.tmpl
```diff
@@ -52,6 +59,9 @@ func (tr *{{ .CRD.Kind }}) GetParameters() (map[string]interface{}, error) {
 	{{- if .CRD.ConfigureExternalName }}
 	{{ .CRD.ConfigureExternalName }}(base, meta.GetExternalName(tr))
 	{{- end }}
+	if err := json.TFParser.Unmarshal(p, &base); err != nil {
+		return nil, err
+	}
 	return base, json.TFParser.Unmarshal(p, &base)
```
muvaf, line 65:
```suggestion
	return base, json.TFParser.Unmarshal(p, &base)
```

pkg/resource/sensitive.go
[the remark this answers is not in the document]
muvaf: Right now it looks like the following:
```go
map[string]string{"master_password": "masterPasswordSecretRef"}
```
In TFState, `master_password` is at the top level in HCL of the resource definition, whereas `spec.forProvider.masterPasswordSecretRef` would have been the fully qualified path for the Crossplane definition. Both work but the latter is correct in the sense that it's ready to be used in all possible contexts whereas if we have only `masterPasswordSecretRef`, we'd need to know the prefix that's needed to use to actually access the data.

pkg/controller/secretclient.go
```diff
@@ -0,0 +1,50 @@
...
+
+	xpv1 "github.com/crossplane/crossplane-runtime/apis/common/v1"
+	"github.com/pkg/errors"
+	v1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/types"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+// APISecretClient is a client for getting k8s secrets
+type APISecretClient struct {
```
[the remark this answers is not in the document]
muvaf: For testing, we have a pretty good [mock utilities](https://github.com/crossplane/crossplane-runtime/blob/master/pkg/test/fake.go) here for generic client and I think it wouldn't be too different if we ended up calling `Get` and access the data in map right where we need it. But it's true that we'd probably need some sort of abstraction in the future to support other backends.

pkg/resource/sensitive.go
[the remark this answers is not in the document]
muvaf: Hmm I didn't think of the case where there is a string as index actually. We'd have to make it not consider strings between `[` and `]` during replace, which could be more complex as you said. Thanks for testing it!

pkg/types/builder.go
muvaf: Could you leave a comment on that issue or open another issue so that we can track it?

pkg/pipeline/templates/terraformed.go.tmpl
turkenh: Ah good catch! I can swear that I deleted them and not sure how they came back, possibly some ctrl+z mistake

pkg/resource/sensitive.go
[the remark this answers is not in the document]
turkenh: Alright, changed like that in the last commit even I don't feel perfect about it since it resulted in some extra string manipulations to trim those prefixes. But sounds good considering it could be used for other contexts.

pkg/types/builder.go
turkenh: https://github.com/crossplane-contrib/terrajet/issues/17

PR #83, opened 2021-10-06
Skip late-initialization of omiempty-tagged maps with zero-length observations

### LINKED ISSUE #82: Late initialize change report returns false positive
### What happened?

While I was testing https://github.com/crossplane-contrib/terrajet/pull/76 , I used `RdsCluster` as async example. After creation, late initializer always reported `true`, meaning that there needs to be a spec update and because of that resource never gets ready since spec update erases changes in the status.

### How can we reproduce it?

I added the [YAML](https://github.com/crossplane-contrib/provider-tf-aws/blob/main/examples/rds/dbcluster.yaml) I used. You can see it never gets ready and if you debug, it's seen that late initializer reports true all the time.

ulucinar:

Fixes #82

I have:

Tested successfully against `provider-tf-aws` with an example `v1alpha1.RdsCluster` resource.

### DIFF
--- a/pkg/resource/lateinit.go
+++ b/pkg/resource/lateinit.go
@@ -89,7 +89,7 @@ func zeroValueJSONOmitEmptyFilter(cName string) ValueFilter {
 			return false
 		case v.IsZero():
 			return true
-		case k == reflect.Slice && v.Len() == 0:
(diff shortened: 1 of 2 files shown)


turkenh (approved):
lgtm!

PR #90, opened 2021-10-07
Generate pointer types for required & optional attributes

### LINKED ISSUE #88: Generate Required Fields with Pointer Types
### What happened?

Currently in Terrajet, we generate fields that correspond to Terraform attributes with a required schema such as [this one](https://github.com/hashicorp/terraform-provider-azurerm/blob/99838548fa435838e2e2f546433bedfc80d54c51/internal/services/containers/kubernetes_cluster_resource.go#L410) with non-pointer types. When we late initialize an instance of the enclosing type ([`network_profile`](https://github.com/hashicorp/terraform-provider-azurerm/blob/99838548fa435838e2e2f546433bedfc80d54c51/internal/services/containers/kubernetes_cluster_resource.go#L400) in this example) with an optional schema, we only late-initialize the optional, hence pointer-typed fields) but not the fields that are marked as required. This is because the late-initialization library only initializes nil-pointers because zero-values of non-pointer types (e.g., an empty string) could be a valid value for that attribute, which should not be late-initialized.

### How can we reproduce it?

When trying #77 with [`provider-tf-azure`](https://github.com/crossplane-contrib/provider-tf-azure/pull/6) on a [v1alpha1.KubernetesCluster](https://github.com/crossplane-contrib/provider-tf-azure/blob/main/examples/kubernetes/kubernetescluster.yaml) resource, first the Kubernetes cluster is provisioned successfully but after the first observation causes a `NetworkPlugin` to be late-initialized without the `NetworkPlugin` field (a required field) being late-initialized to one of the allowed values (`kubenet` or `azure`) although the Terraform state has it properly set to `kubenet`.

ulucinar:

Fixes #88

I have:

Example struct fields generated before this change in `provider-tf-azure` for `v1alpha1.KubernetesCluster` MR:
image

And with this change, the type of the field corresponding to the required `network_plugin` attribute becomes:
image

Successfully tested (provisioned/deprovisioned) `provider-tf-azure` `KubernetesCluster`, `PostgresqlServer` and `VirtualNetwork` resources (with custom configurations) generated with this PR.

### DIFF
--- a/pkg/pipeline/templates/terraformed.go.tmpl
+++ b/pkg/pipeline/templates/terraformed.go.tmpl
@@ -79,8 +79,7 @@ func (tr *{{ .CRD.Kind }}) LateInitialize(attrs []byte) (bool, error) {
 	if err := json.TFParser.Unmarshal(attrs, params); err != nil {
 		return false, errors.Wrap(err, "failed to unmarshal Terraform state parameters for late-initialization")
 	}
-	opts := []resource.GenericLateInitializerOption{resource.WithZeroValueJSONOmitEmptyFilter(resource.CNameWildcard),
-	    resource.WithZeroElemPtrFilter(resource.CNameWildcard)}
+	opts := []resource.GenericLateInitializerOption{resource.WithZeroValueJSONOmitEmptyFilter(resource.CNameWildcard)}
 	{{ range .LateInitializer.IgnoredFields -}}
 	    opts = append(opts, resource.WithNameFilter("{{ . }}"))
 	{{ end }}
--- a/pkg/types/builder.go
+++ b/pkg/types/builder.go
@@ -22,11 +22,10 @@ import (
 	"go/types"
 	"sort"
 
+	"github.com/crossplane/crossplane-runtime/pkg/fieldpath"
 	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
 	twtypes "github.com/muvaf/typewriter/pkg/types"
 	"github.com/pkg/errors"
-
-	"github.com/crossplane/crossplane-runtime/pkg/fieldpath"
 
 	"github.com/crossplane-contrib/terrajet/pkg/config"
 	"github.com/crossplane-contrib/terrajet/pkg/types/comments"


pkg/types/builder.go
```diff
@@ -22,11 +22,12 @@ import (
 	"go/types"
 	"sort"
 
-	"github.com/crossplane/crossplane-runtime/pkg/fieldpath"
 	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
 	twtypes "github.com/muvaf/typewriter/pkg/types"
 	"github.com/pkg/errors"
 
+	"github.com/crossplane/crossplane-runtime/pkg/fieldpath"
```
muvaf, line 29: We group these with the upper block since it's in a different github org, at least for now.

ulucinar: Thanks @muvaf. That's my IDE kicking in.

pkg/types/builder.go
```diff
@@ -193,25 +193,13 @@ func (g *Builder) buildResource(res *schema.Resource, cfg *config.Resource, tfPa
 func (g *Builder) buildSchema(sch *schema.Schema, cfg *config.Resource, tfPath []string, xpPath []string, names []string) (types.Type, error) { // nolint:gocyclo
 	switch sch.Type {
```
muvaf, line 194: It'd be great to have a comment here explaining why we diverge from the pattern of making required fields non-pointer type.

pkg/pipeline/templates/terraformed.go.tmpl
```diff
@@ -79,8 +79,7 @@ func (tr *{{ .CRD.Kind }}) LateInitialize(attrs []byte) (bool, error) {
 	if err := json.TFParser.Unmarshal(attrs, params); err != nil {
 		return false, errors.Wrap(err, "failed to unmarshal Terraform state parameters for late-initialization")
 	}
-	opts := []resource.GenericLateInitializerOption{resource.WithZeroValueJSONOmitEmptyFilter(resource.CNameWildcard),
-	    resource.WithZeroElemPtrFilter(resource.CNameWildcard)}
+	opts := []resource.GenericLateInitializerOption{resource.WithZeroValueJSONOmitEmptyFilter(resource.CNameWildcard)}
```
muvaf, line 82: Why don't we need `resource.WithZeroElemPtrFilter(resource.CNameWildcard)` anymore?

muvaf (approved):
LGTM! Thank you @ulucinar !

PR #94, opened 2021-10-08
Fix invalid key syntax for paths with indexes

### LINKED ISSUE #93: Cannot use Paved path as a key in Kubernetes secret
### What happened?

While trying to store sensitive data from a `v1alpha1.KubernetesCluster` in provider-tf-azure, after setting the `writeConnectionSecretToRef` field, I observed the following logs:

```
2021-10-07T23:33:55.924+0300    DEBUG   controller-runtime.manager.events       Warning {"object": {"kind":"KubernetesCluster","name":"example","uid":"1a8fbb21-9823-4869-8c3a-37b76bd9e1bc","apiVersion":"kubernetes.azure.tf.crossplane.io/v1alpha1","resourceVersion":"643827"}, "reason": "CannotPublishConnectionDetails", "message": "cannot create or update connection secret: cannot create object: Secret \"test-k8s\" is invalid: [data[kube_config[0].password]: Invalid value: \"kube_config[0].password\": a valid config key must consist of alphanumeric characters, '-', '_' or '.' (e.g. 'key.name',  or 'KEY_NAME',  or 'key-name', regex used for validation is '[-._a-zA-Z0-9]+'), data[kube_config[0].client_certificate]: Invalid value: \"kube_config[0].client_certificate\": a valid config key must consist of alphanumeric characters, '-', '_' or '.' (e.g. 'key.name',  or 'KEY_NAME',  or 'key-name', regex used for validation is '[-._a-zA-Z0-9]+'), data[kube_config[0].client_key]: Invalid value: \"kube_config[0].client_key\": a valid config key must consist of alphanumeric characters, '-', '_' or '.' (e.g. 'key.name',  or 'KEY_NAME',  or 'key-name', regex used for validation is '[-._a-zA-Z0-9]+'), data[kube_config[0].cluster_ca_certificate]: Invalid value: \"kube_config[0].cluster_ca_certificate\": a valid config key must consist of alphanumeric characters, '-', '_' or '.' (e.g. 'key.name',  or 'KEY_NAME',  or 'key-name', regex used for validation is '[-._a-zA-Z0-9]+')]"}
```

### How can we reproduce it?

Using provider-tf-azure (from a local build):
```
apiVersion: kubernetes.azure.tf.crossplane.io/v1alpha1
kind: KubernetesCluster
metadata:
  name: example
spec:
  writeConnectionSecretToRef:
    name: example-kubeconfig
    namespace: crossplane-system
  forProvider:
    name: "kubernetesCluster1"
    location: "East US"
    resourceGroupName: "example-rg"
    dnsPrefix: "exampleaks1"
    defaultNodePool:
      - name: default
        nodeCount: 1
        vmSize: "Standard_D2_v2"
    identity:
      - type: "SystemAssigned"
    tags:
      purpose: example
  providerConfigRef:
    name: example

```

turkenh:

As @ulucinar revealed in #93 Kubernetes secret keys use a limited character set to validate secret keys which does not include `[` and `]` which we were using in keys for indexed paths.

We need to store this information in connection details secret but still have to be able to construct it back without losing information to build the tfstate.

This PR fixed the problem by converting:

`kubeconfig[0].password`

to 

`kubeconfig.0.password`

There is one edge which could break this, that is, keys containing dots inside (e.g. `ip_addresses[crossplane.io]`). We are not sure how common is this among the sensitive terraform attributes but this case also handled as follows:

`ip_addresses[crossplane.io]`

to 

`ip_addresses...crossplane.io...`

Fixes #93

I have:

Unit tests
I also verified k8s secret could be created with keys three dots like `abc...my.cool.key...` which corresponds to `abc[my.cool.key]`

### DIFF
--- a/pkg/resource/sensitive.go
+++ b/pkg/resource/sensitive.go
@@ -18,6 +18,8 @@ package resource
 
 import (
 	"context"
+	"fmt"
+	"regexp"
 	"strings"
 
 	kerrors "k8s.io/apimachinery/pkg/api/errors"
@@ -33,6 +35,16 @@ const (
 
 	errFmtCannotGetStringForFieldPath = "cannot not get a string for fieldpath %q"
 )
+
+var reEndsWithIndex *regexp.Regexp
+var reMiddleIndex *regexp.Regexp
+var reInsideThreeDotsBlock *regexp.Regexp
+
+func init() {
+	reEndsWithIndex = regexp.MustCompile(`\.(\d+?)$`)
+	reMiddleIndex = regexp.MustCompile(`\.(\d+?)\.`)
+	reInsideThreeDotsBlock = regexp.MustCompile(`\.\.\.(.*?)\.\.\.`)
+}
 
 // SecretClient is the client to get sensitive data from kubernetes secrets
 //go:generate go run github.com/golang/mock/mockgen -copyright_file ../../hack/boilerplate.txt -destination ./fake/mocks/mock.go -package mocks github.com/crossplane-contrib/terrajet/pkg/resource SecretClient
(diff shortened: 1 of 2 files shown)


pkg/resource/sensitive.go
```diff
@@ -51,17 +63,21 @@ func GetConnectionDetails(from map[string]interface{}, mapping map[string]string
...
 		}
 
-		for _, s := range segments {
-			v, err := paved.GetString(s)
+		for _, fp := range fieldPaths {
+			v, err := paved.GetString(fp)
+			if err != nil {
+				return nil, errors.Wrapf(err, errFmtCannotGetStringForFieldPath, fp)
+			}
+			k, err := fieldPathToSecretKey(fp)
```
muvaf, line 76: Can you add a comment here explaining why we need to do this conversion?

muvaf (approved):
LGTM!

pkg/resource/sensitive.go
```diff
@@ -199,3 +219,37 @@ func expandedFor(expanded fieldpath.Segments, withWildcard fieldpath.Segments) b
...
+	if err != nil {
+		return "", errors.Wrapf(err, "cannot parse %q as fieldpath", s)
+	}
+
+	var b strings.Builder
+	for _, s := range sg {
+		switch s.Type {
+		case fieldpath.SegmentField:
+			if strings.ContainsRune(s.Field, '.') {
+				b.WriteString(fmt.Sprintf("...%s...", s.Field))
```
muvaf, line 245: I don't love having three dots here but seems like we don't have another choice.

PR #103, opened 2021-10-11
Callback for updating status with last operation error

muvaf:

Currently, when an error occurs in async calls, we'd reconcile after a whole minute to see the error, just like native async controllers like AWS RDS Instance. However, we can use the existing callback mechanism to set the error condition and that'd get us a re-queue if there is a change, which will then allow us process the resulting state.

I have:

Manually with RDS cluster resource and EC2 Instance.

### DIFF
--- a/pkg/controller/api.go
+++ b/pkg/controller/api.go
@@ -32,6 +32,10 @@ import (
 	"github.com/crossplane-contrib/terrajet/pkg/terraform"
 )
 
+const (
+	errGet = "cannot get resource"
+)
+
 // APISecretClient is a client for getting k8s secrets
 type APISecretClient struct {
 	kube client.Client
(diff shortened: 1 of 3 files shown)


pkg/controller/api.go
```diff
@@ -0,0 +1,101 @@
...
+	d, err := a.GetSecretData(ctx, &sel.SecretReference)
+	if err != nil {
+		return nil, errors.Wrap(err, "cannot get secret data")
+	}
+	return d[sel.Key], err
+}
+
+// NewAPICallbacks returns a new APICallbacks.
+func NewAPICallbacks(m ctrl.Manager, of xpresource.ManagedKind) *APICallbacks {
+	nt := func() resource.Terraformed {
```
muvaf, line 60: I opted for this pattern so that we need to keep only the `metadata.name` of the resource rather than the whole object, which is different than the earlier implementation.

PR #121, opened 2021-10-19
Add custom config option for defining custom connection keys

### LINKED ISSUE #86: Ability to construct custom connection detail keys

Currently, all fields marked as sensitive in Terraform schema were added as a connection detail key and were hidden from the schema (i.e. dropped if in status, converted to `secretref` if in spec). We are using terraform field names as keys which would also be helpful in terms of documenting what a connection detail secret includes as keys.

However, there could be some cases where it makes sense to build a custom key specific to resource. For example, if a managed Kubernetes cluster returns different parts of a kubeconfig as separate keys, it would be convenient to construct a kubeconfig using them and store with `kubeconfig` key.

We should expose a custom configuration which allows to define additional keys for connection details by getting a function similar to external name config, for example:

`func CustomConnectionConstructor(attr map[string]interface{}) map[string][]byte`

turkenh:

This PR adds a custom configuration option for connection detail keys where provider developers could provide a go function that takes Terraform state attributes and returns a map containing custom keys/values that they want to see in the connection details secret as. With this interface, provider developers could build a connection key in any format they want, using available terraform attributes, e.g. SQL connection string as a single key, or add a nonsensitive field as connection key, like `endpoint`.

In this PR, we are also adding `attribute_` prefix for keys of sensitive terraform fields. We needed this prefix to ensure that they are not overridden by any custom connection key configured which would break our ability to build `tfstate` back. Considering Terraform already lists them under [Attributes](https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/iam_access_key#attributes-reference), this should be intuitive.

Fixes #86 

I have:

With provider-tf-aws using https://github.com/crossplane-contrib/provider-tf-aws/pull/119/files

1. Create a iam `AccessKey` 
2. Check its connection secret

Observed a secret like where aws_access_key_id and aws_secret_access_key are custom keys:

```
apiVersion: v1
data:
  attribute_id: QUtJQVk0QUZUVFNFNDI2TlhKS0I=
  attribute_secret: ABCxyzRedacted==
  attribute_ses_smtp_password_v4: QQ00REDACTED==
  aws_access_key_id: QUtJQVk0QUZUVFNFNDI2TlhKS0I=
  aws_secret_access_key: ABCxyzRedacted==
kind: Secret
```

### DIFF
--- a/pkg/config/resource.go
+++ b/pkg/config/resource.go
@@ -83,8 +83,8 @@ type References map[string]Reference
 
 // Sensitive represents configurations to handle sensitive information
 type Sensitive struct {
-	// CustomKeysFunctionPath is the path for function adding custom connection
-	// details keys
+	// AdditionalConnectionDetailsFn is the path for function adding additional
+	// connection details keys
 	AdditionalConnectionDetailsFn AdditionalConnectionDetailsFn
 
 	fieldPaths map[string]string
--- a/pkg/controller/external_test.go
+++ b/pkg/controller/external_test.go
@@ -20,8 +20,6 @@ import (
 	"context"
 	"testing"
 
-	"github.com/crossplane-contrib/terrajet/pkg/config"
-
 	xpv1 "github.com/crossplane/crossplane-runtime/apis/common/v1"
 	"github.com/crossplane/crossplane-runtime/pkg/reconciler/managed"
 	xpresource "github.com/crossplane/crossplane-runtime/pkg/resource"
@@ -31,6 +29,7 @@ import (
 	"github.com/pkg/errors"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 
+	"github.com/crossplane-contrib/terrajet/pkg/config"
 	"github.com/crossplane-contrib/terrajet/pkg/resource"
 	"github.com/crossplane-contrib/terrajet/pkg/resource/fake"
 	"github.com/crossplane-contrib/terrajet/pkg/resource/json"
--- a/pkg/resource/lateinit.go
+++ b/pkg/resource/lateinit.go
@@ -23,7 +23,6 @@ import (
 	"strings"
 
 	xpmeta "github.com/crossplane/crossplane-runtime/pkg/meta"
-
 	"github.com/pkg/errors"
 )
 


pkg/pipeline/templates/terraformed.go.tmpl
```diff
@@ -24,13 +24,22 @@ func (tr *{{ .CRD.Kind }}) GetTerraformResourceIDField() string {
...
-  {{- if .SensitiveFields }}
-  return map[string]string{ {{range $k, $v := .SensitiveFields}}"{{ $k }}": "{{ $v}}", {{end}} }
+  {{- if .Sensitive.Fields }}
+  return map[string]string{ {{range $k, $v := .Sensitive.Fields}}"{{ $k }}": "{{ $v}}", {{end}} }
   {{- else }}
   return nil
   {{- end }}
 }
 
+// GetCustomConnectionKeys for this {{ .CRD.Kind }}
```
muvaf, line 34: Do you think we can follow an approach similar to https://github.com/crossplane-contrib/terrajet/pull/124 ?

pkg/resource/sensitive.go
```diff
@@ -31,11 +31,18 @@ import (
...
+	errCannotExpandWildcards          = "cannot expand wildcards"
 	errFmtCannotGetStringForFieldPath = "cannot not get a string for fieldpath %q"
 )
 
+const (
+	// PrefixAttribute used to prefix connection detail keys for sensitive
+	// Terraform attributes. We need this prefix to ensure that they are not
+	// overridden by any custom connection key configured which would break
+	// our ability to build tfstate back.
+	PrefixAttribute = "attribute_"
```
muvaf, line 43: nitpick: would it be possible to use `attributes.` to make it similar to how it appears on tfstate json?

pkg/pipeline/templates/terraformed.go.tmpl
turkenh: Yeah, definetely. I'll rebase on top of it so that it can go first.

pkg/resource/sensitive.go
turkenh: yes, good point!

pkg/resource/fake/terraformed.go
```diff
@@ -40,6 +41,11 @@ func (o *Observable) SetObservation(data map[string]interface{}) error {
 	return nil
 }
 
+// GetCustomConnectionKeys is a mock
+func (o *Observable) GetCustomConnectionKeys(_ map[string]interface{}) (map[string][]byte, error) {
```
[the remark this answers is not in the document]
turkenh: Changed the name as you suggested but ended up using the same signature because I don't want to open a door where existing keys are overridden by custom keys defined by the provider developer. Right now, I check it and return an explicit error that this is not supported and this could be done in a more straightforward way with this signature.

turkenh:
@muvaf this should be ready to go!

pkg/resource/fake/terraformed.go
```diff
@@ -26,7 +26,8 @@ import (
 
 // Observable is mock Observable.
 type Observable struct {
-	Observation map[string]interface{}
+	Observation          map[string]interface{}
+	CustomConnectionKeys map[string][]byte
```
muvaf, line 30:
```suggestion
	AdditionalConnectionDetails map[string][]byte
```

pkg/resource/sensitive.go
```diff
@@ -31,11 +31,18 @@ import (
...
+	errCannotExpandWildcards          = "cannot expand wildcards"
 	errFmtCannotGetStringForFieldPath = "cannot not get a string for fieldpath %q"
 )
 
+const (
+	// PrefixAttribute used to prefix connection detail keys for sensitive
+	// Terraform attributes. We need this prefix to ensure that they are not
+	// overridden by any custom connection key configured which would break
+	// our ability to build tfstate back.
+	PrefixAttribute = "attribute."
```
muvaf, line 43: Given that now we return error if user tried to override an attribute key, do we still need this?

pkg/controller/external.go
```diff
@@ -291,3 +284,39 @@ func CriticalAnnotationsCallback(kube client.Client, tr resource.Terraformed) te
 		return errors.Wrap(c.UpdateCriticalAnnotations(ctx, tr), "cannot update critical annotations")
 	}
 }
+
+func getConnectionDetails(attr map[string]interface{}, tr resource.Terraformed, cfg config.Resource) (managed.ConnectionDetails, error) {
```
muvaf, line 288: I'm not sure if we should have `getConnectionDetails` and `lateInitializeAnnotations` functions here. They seem like resource operations, hence could go to `resource` package. What do you think?

pkg/controller/external.go
```diff
@@ -291,3 +284,39 @@ func CriticalAnnotationsCallback(kube client.Client, tr resource.Terraformed) te
...
+	if cfg.Sensitive.CustomKeysFn != nil {
+		if custom, err = cfg.Sensitive.CustomKeysFn(attr); err != nil {
+			return nil, errors.Wrap(err, "cannot get custom connection keys")
+		}
+	}
+
+	if conn == nil {
+		// if there is no sensitive attributes but still some custom connection
+		// keys, e.g. endpoint
+		conn = map[string][]byte{}
```
muvaf, line 305: nitpick
```suggestion
	var custom map[string][]byte
	// TODO(turkenh): Once we have automatic defaulting, remove this if check.
	if cfg.Sensitive.CustomKeysFn != nil {
		if custom, err = cfg.Sensitive.CustomKeysFn(attr); err != nil {
			return nil, errors.Wrap(err, "cannot get custom connection keys")
		}
	}

	if conn == nil {
		// if there is no sensitive attributes but still some custom connection
		// keys, e.g. endpoint
		return custom, nil
```

muvaf (commented):
Thanks @turkenh ! I love how easy we're able to hook up the custom config to the secret flow we have in place here.

Issue #123, opened 2021-10-19
Call external name configurator in controller


muvaf:
### What problem are you facing?

Currently, the custom external name config function is called in `apis/group/v1alpha1/zz_.._terraformed.go`. So, the type package of the CRD depends on the package that the external name configuration lives.

We used to have that custom func in types package but we're moving towards having all custom code in a separate folder like top-level `config`. So, the problem we face now is that we cannot use any CRD types in `config` folder because it'd result in circular dependency.

### How could Terrajet help solve your problem?

Instead of calling the function in types package, we can have the controller call it, hence the dependency tree would look like `controller -> config -> apis`. This would enable us to:
* Use CRD structs in custom reference extractors, hence not have to marshal objects into unstructured in those functions.
* `config` becomes importable by `cmd/provider`, hence we'd be able to reference functions directly instead of giving their type path.

The last point I believe we discussed with @turkenh a while ago. We can't make reference config use strong-typed functions, hence wanted the rest to use func path as well but there wasn't many config points back then. Now that we have more configuration interfaces, I think it's worth to keep ref resolver use path and the rest use strong-typed structs.

turkenh:
Sounds good @muvaf, I agree that directly providing function instead of their paths would be a nice improvement 👍 

> Use CRD structs in custom reference extractors, hence not have to marshal objects into unstructured in those functions.

Actually, this was helpful because we would use some generic functions across multiple resources instead of writing the same for all of them.  For example, [ARNExtractor()](https://github.com/crossplane-contrib/provider-tf-aws/blob/main/config/common/common.go#L46) in provider-tf-aws was used around 15-20 times, and assuming we would otherwise need to write about 5-10 strong typed functions, I am not sure if the performance improvement we get by not marshaling is really significant that requires us to do so.

PR #124, opened 2021-10-19
external name: use function in the config directly instead of generating the call

muvaf:

We used to print the function call in terraformed.go, but with the recent changes and learnings it seems like calling the function directly on the config object during runtime will enable us to implement more streamlined features. See details in https://github.com/crossplane-contrib/terrajet/issues/123

See https://github.com/crossplane-contrib/provider-tf-aws/pull/122 for its application. Once merged, I'll do the same changes in Azure as well.

Fixes https://github.com/crossplane-contrib/terrajet/issues/123

I have:

Manually with AWS VPC.

### DIFF
--- a/pkg/config/common.go
+++ b/pkg/config/common.go
@@ -1,0 +1,40 @@
+/*
+Copyright 2021 The Crossplane Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package config
+
+// Common ExternalName configurations.
+var (
+	// NameAsIdentifier uses "name" field in the arguments as the identifier of
+	// the resource.
+	NameAsIdentifier = ExternalName{
+		SetIdentifierArgumentFn: func(base map[string]interface{}, name string) {
+			base["name"] = name
+		},
+		OmittedFields: []string{
+			"name",
+			"name_prefix",
+		},
+	}
+
+	// IdentifierFromProvider is used in resources whose identifier is assigned by
+	// the remote client, such as AWS VPC where it gets an identifier like
+	// vpc-2213das instead of letting user choose a name.
+	IdentifierFromProvider = ExternalName{
+		SetIdentifierArgumentFn: NopSetIdentifierArgument,
+		DisableNameInitializer:  true,
+	}
+)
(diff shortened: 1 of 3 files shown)


pkg/config/resource.go
```diff
@@ -43,6 +43,9 @@ func NewResource(version, kind, terraformResourceType string, opts ...ResourceOp
 		Kind:                  kind,
 		TerraformResourceType: terraformResourceType,
 		TerraformIDFieldName:  "id",
+		ExternalName: ExternalName{
+			ConfigureFn: NopConfigureWithName,
```
turkenh, line 47: I had a hard time understanding the purpose of this option in the beginning because it sounds like we are configuring external name, like providing a function to build external name. However, it is almost the opposite.
Do you think if we could find a more intuitive name this? Something like `ExternalName.SetTerraformArgsFn` or at least `ExternalName.ConsumeFn`.

turkenh (approved):

pkg/config/resource.go
muvaf: Hmm how does `SetNameAttributeFn` sound?

turkenh: Yeah sounds better. Considering we need to set fields like [cluster_identifier](https://github.com/crossplane-contrib/provider-tf-aws/blob/5509c10d768622c3631615947cf7b18086a58aa3/config/rds/config.go#L32), I didn't feel great about having _Name_ there. May be `SetAttributesFn` or `SetArgumentsFn` ? The reason why I suggest Argument was, we would always need to set field(s) from [the Arguments section](https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/vpc#argument-reference) in Terraform docs.

muvaf: I see. I don't want it as generic as `SetAttributes` though. So I changed it to `SetIdentifierArgumentFn` which is more precise.

turkenh: Sounds great, thanks!

pkg/config/common.go
```diff
@@ -0,0 +1,40 @@
...
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package config
+
+// Common ExternalName configurations.
+var (
```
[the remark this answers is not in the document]
turkenh: Sounds good, also we could expose this as a provider-level configuration (e.g. default external name config) that could still be overridden by specific resources. For example, for azure, we can just set `provider.defaultExternalNameConfig = IdentifierFromProvider` and done.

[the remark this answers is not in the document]
muvaf: Yeah but if we're going to have that for loop anyway, it could be simpler to set it every time so that we'd worry about only a single field in a single config object instead of checking the default if it doesn't exist.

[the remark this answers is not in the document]
turkenh: Maybe we can find a way not to have that loop :)

PR #125, opened 2021-10-20
Generate code that handles Crossplane's `enabledAPIs` for terrajet-based providers

ulucinar:

Related Crossplane issue: https://github.com/crossplane/crossplane/issues/2122

Please see https://github.com/crossplane/crossplane/issues/2122 for a detailed account.

I have:

Please see https://github.com/crossplane-contrib/provider-tf-azure/pull/74 for a detailed account on testing via `provider-tf-azure`.

### DIFF
--- a/pkg/pipeline/templates/controller.go.tmpl
+++ b/pkg/pipeline/templates/controller.go.tmpl
@@ -23,7 +23,15 @@ import (
 )
 
 // Setup adds a controller that reconciles {{ .CRD.Kind }} managed resources.
-func Setup(mgr ctrl.Manager, l logging.Logger, rl workqueue.RateLimiter, s terraform.SetupFn, ws *terraform.WorkspaceStore, concurrency int) error {
+func Setup(mgr ctrl.Manager, l logging.Logger, rl workqueue.RateLimiter, s terraform.SetupFn, ws *terraform.WorkspaceStore, concurrency int, enabledAPIs []string) error {
+    enabled, err := tjcontroller.IsAPIEnabled({{ .TypePackageAlias }}{{ .CRD.Kind }}GroupVersionKind, enabledAPIs)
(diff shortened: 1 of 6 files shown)


pkg/pipeline/templates/controller.go.tmpl
```diff
@@ -23,7 +23,15 @@ import (
 )
 
 // Setup adds a controller that reconciles {{ .CRD.Kind }} managed resources.
-func Setup(mgr ctrl.Manager, l logging.Logger, rl workqueue.RateLimiter, s terraform.SetupFn, ws *terraform.WorkspaceStore, concurrency int) error {
+func Setup(mgr ctrl.Manager, l logging.Logger, rl workqueue.RateLimiter, s terraform.SetupFn, ws *terraform.WorkspaceStore, concurrency int, enabledAPIs []string) error {
```
muvaf, line 26: I think we'll need a similar logic in `AddToScheme` as well since we can't register schemes of CRDs that are not installed.

Would it be possible to handle all this enable/disable logic in `cmd/provider/main.go`? i.e. maybe accumulate all `AddToScheme` and `Setup` function calls in a map instead of calling them in `zz_register.go` and `zz_setup.go`, then do the filtering in `cmd/provider/main.go` and only call the ones that are enabled? We could possibly contain the whole enable/disable APIs logic in a single place this way.

Since this will affect all providers, at a later iteration, we can think of having some logic in crossplane-runtime to handle all this in a streamlined way.

muvaf:
@ulucinar Now that the upstream fixes [made it](https://github.com/crossplane/crossplane/issues/2649#issuecomment-966397577) to patch releases, do you think we should close this PR?

ulucinar:
Closing this PR we have the upstream K8s patches merged.

PR #130, opened 2021-11-01
Generating a Provider guide

### LINKED ISSUE #42: Guide to generate a new TF provider

Users don't know how to generate a TF provider.

We can have a step-by-step guide for generating a new TF based provider using terrajet.

turkenh:

This PR adds two document:

- Generating a Provider using Terrajet
- Configuring a Resource

Fixes #42 

I have:

Run the steps end to end and verify it works.


docs/generating-a-provider.md
```diff
@@ -0,0 +1,391 @@
...
+[comment]: <> (References)
+
+[Terraform GitHub provider]: https://registry.terraform.io/providers/integrations/github/latest/docs
+[provider-tf-template]: https://github.com/crossplane-contrib/provider-tf-template
+[Terraform documentation for provider configuration]: https://registry.terraform.io/providers/integrations/github/latest/docs#argument-reference
+[`github` directory]: https://github.com/integrations/terraform-provider-github/tree/main/github
+[this line in controller Dockerfile]: https://github.com/crossplane-contrib/provider-tf-template/blob/70f485a3f227d30e7eaac43aaf42348f7e930232/cluster/images/provider-tf-template-controller/Dockerfile#L25
+[fork]: https://github.com/turkenh/terraform-provider-github
+[terraform-provider-github]: https://github.com/integrations/terraform-provider-github
+[terraform-plugin-sdk]: github.com/hashicorp/terraform-plugin-sdk
```
sergenyalcin, line 390: Hi @turkenh, this link does not work. We may consider to fix this.
```suggestion
[terraform-plugin-sdk]: https://github.com/hashicorp/terraform-plugin-sdk
```

sergenyalcin:
I tried creating a provider and configuring a resource in the provider, following the steps in the documentation. The document was clear to me.

turkenh:
> I tried creating a provider and configuring a resource in the provider, following the steps in the documentation. I completed the tests without any problems. These documents are clear to me.

Thanks for trying it out @sergenyalcin, much appreciated!

docs/generating-a-provider.md
```diff
@@ -0,0 +1,391 @@
...
+
+    2. Replace all occurrences of `template` with your provider name:
+
+        ```bash
+        git grep -l 'template' -- './*' ':!build/**' ':!go.sum' | xargs sed -i.bak "s/template/${ProviderNameLower}/g"
+        git grep -l 'Template' -- './*' ':!build/**' ':!go.sum' | xargs sed -i.bak "s/Template/${ProviderNameUpper}/g"
+        # Clean up the .bak files created by sed
+        git clean -fd
+   
+        mv "internal/clients/template.go" "internal/clients/${ProviderNameLower}.go"
```
displague, line 31: git mv?

docs/generating-a-provider.md
```diff
@@ -0,0 +1,391 @@
...
+
+4. Configure your repo for the Terraform provider binary and schema:
+
+    1. Update the following variables in `Makefile` for Terraform Provider:
+
+        ```makefile
+        export TERRAFORM_PROVIDER_SOURCE := integrations/github
+        export TERRAFORM_PROVIDER_VERSION := 4.17.0
+        export TERRAFORM_PROVIDER_DOWNLOAD_NAME := terraform-provider-github
+        export TERRAFORM_PROVIDER_DOWNLOAD_URL_PREFIX := https://releases.hashicorp.com/terraform-provider-github/4.17.0
```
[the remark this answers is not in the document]
displague: Perhaps not - https://releases.hashicorp.com/ does not include a majority of independent providers.
https://github.com/equinix/terraform-provider-metal/releases/download/v3.2.0/terraform-provider-metal_3.2.0_linux_amd64.zip is where my provider binary resides, for example.

docs/generating-a-provider.md
```diff
@@ -0,0 +1,391 @@
...
+       -       tf "github.com/hashicorp/terraform-provider-hashicups/hashicups"
+       +       tf "github.com/turkenh/terraform-provider-github/v4/github"
+       )
+  
+       const resourcePrefix = "github"
+       ```
+
+       Run:
+       ```bash
+       go mod tidy
```
displague, line 69: It may be helpful to note that providers that are >v1 without being compatible with https://go.dev/blog/v2-go-modules will need to set the version of their provider with the `+` format (I don't know what to call this format and I don't know how to get it back from a command - I always dig up and truncate the commit sha and timestamp by hand).

docs/generating-a-provider.md
```diff
@@ -0,0 +1,391 @@
...
+In this guide, we will generate a Crossplane provider based on an existing
+Terraform provider using Terrajet.
+
+We have chosen [Terraform GitHub provider] as an example, but the process will
+be quite similar for any other Terraform provider.
+
+## Generate
+
+1. Generate a GitHub repository for the Crossplane provider by hitting the
+   "**Use this template**" button in [provider-tf-template] repository.
```
zzxwill, line 12: Will you add a link for `provider-tf-template` later?

zzxwill, line 12: Is the repo `[provider-tf-template]` ready now? I'd like to go through it.

zzxwill: Sorry I got the answer in the references https://github.com/crossplane-contrib/provider-tf-template. Thanks.

docs/generating-a-provider.md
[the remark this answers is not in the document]
zzxwill: Maybe we should also set the version for Terraform binary here as a provider can only work with some specified binary versions. For Alibaba cloud provider, 
>From version 1.50.0, the provider starts to support Terraform 0.12.x.

https://registry.terraform.io/providers/aliyun/alicloud/latest/docs

The version of Terraform binary can help build suitable terraform images for .tf execution.

docs/generating-a-provider.md
```diff
@@ -0,0 +1,391 @@
...
+   resources as follows:
+
+   ```bash
+   # Create custom configuration directory for whole repository group
+   mkdir config/repository
+   # Create custom configuration directory for whole branch group
+   mkdir config/branch
+   ```
+
+   ```bash
```
muvaf, line 160:
```suggestion
   ```go
```

Because they won't write the exact code, seeing it in command format won't really help as much. So, I'd suggest having YAMLs and Go snippets as is to get syntax highlighting and get users to copy-paste them in their editors instead of running commands. The same doesn't go for the generic find&replace commands we have at the top for example, because they will run them regardless of the provider they're generating.

docs/generating-a-provider.md
```diff
@@ -0,0 +1,391 @@
...
+   package branch
+
+   import "github.com/crossplane-contrib/terrajet/pkg/config"
+
+   func Customize(p *config.Provider) {
+       p.AddResourceConfigurator("github_branch", func(r *config.Resource) {
+   
+           // we need to override the default group that terrajet generated for
+           // this resource, which would be "github" 
+           r.Group = "branch"
```
muvaf, line 188:
```suggestion
           r.Group = "repos"
```

In the original [Github API](https://docs.github.com/en/rest/reference/repos#branches), that's the group used for both `repository` and `branch` resources, which is what we're following in [native provider-github](https://github.com/crossplane-contrib/provider-github/blob/main/apis/organizations/v1alpha1/register.go#L28), too. As I mentioned in another PR, I think we should tell people to follow the API instead of Terraform grouping so that migration/merge scenarios are less-breaking.

docs/generating-a-provider.md
```diff
@@ -0,0 +1,391 @@
...
+   ```go
+   pc := tjconfig.NewProvider(resourceMap, resourcePrefix, modulePath,
+       tjconfig.WithDefaultResourceFn(defaultResourceFn),
+       tjconfig.WithIncludeList([]string{
+           "github_repository$",
+           "github_branch$",
+       }))
+   ```
+
+7. Finally, we would need to add some custom configurations for these two
```
muvaf, line 150: Because the custom config doc goes into great detail, what do you think leaving provider generation part end here? i.e. use `.*` to generate all and then let them know they can now provide custom configs by following that other guide.

docs/configuring-a-resource.md
```diff
@@ -0,0 +1,410 @@
...
+However, there are still couple of information that requires some input
+configuration which could easily be provided by checking the Terraform
+documentation of the resource:
+
+- [External name]
+- [Cross Resource Referencing]
+- [Additional Sensitive Fields and Custom Connection Details]
+- [Late Initialization Behavior]
+
+### External Name
```
muvaf, line 15: I think this section needs an update because of https://github.com/crossplane-contrib/terrajet/pull/132

docs/configuring-a-resource.md
```diff
@@ -0,0 +1,410 @@
...
+    // more than the actual name of the resource, like subscription ID or region
+    // etc. which is unlikely to be included in metadata.name
+    DisableNameInitializer bool
+}
+```
+
+Comments explain the purpose of each field but let's clarify further with some
+examples.
+
+Checking the [import section of aws_vpc], we see that this resource is being
```
muvaf, line 56: nitpick: I'd prefer to start with the usual case and continue with the `IdentifierFromProvider` case so that readers make the effort to do it properly and only if they fail, they continue with modifications.

docs/configuring-a-resource.md
```diff
@@ -0,0 +1,410 @@
...
+}
+```
+
+Please note, you can always check resource configurations of existing Providers
+as further examples under `config/<group>/config.go`.
+
+### Cross Resource Referencing
+
+Crossplane uses cross resource referencing to [handle dependencies] between
+managed resources. For example, if you have an iam User defined as a Crossplane
```
muvaf, line 149:
```suggestion
managed resources. For example, if you have an IAM User defined as a Crossplane
```

docs/configuring-a-resource.md
```diff
@@ -0,0 +1,410 @@
...
+    userRef:
+      name: sample-user
+  writeConnectionSecretToRef:
+    name: sample-access-key-secret
+    namespace: crossplane-system
+```
+
+Historically, reference resolution method were written by hand which requires
+some effort, however, with the latest Crossplane code generation tooling, it is
+now possible to [generate reference resolution methods] by just adding some
```
muvaf, line 180: nitpick: not sure how it used to be recently is relevant since this guide will live months/years.

docs/configuring-a-resource.md
```diff
@@ -0,0 +1,410 @@
...
+```
+
+### Additional Sensitive Fields and Custom Connection Details
+
+Crossplane stores sensitive information of a managed resource in a Kubernetes
+secret, together with some additional fields that would help consumption of the
+resource, a.k.a. [connection details].
+
+In Terrajet, we already [handle sensitive fields] that are marked as sensitive
+in Terraform schema and no further action required for them. However, we still
```
muvaf, line 251: I think we should tell them that if the field is marked as sensitive it's automatically removed from CRD and published only in the connection details secret.

docs/configuring-a-resource.md
```diff
@@ -0,0 +1,410 @@
...
+have some custom configuration API that would allow marking additional fields as
+sensitive (e.g. just if we encounter a field that is not marked properly) and
```
muvaf, line 253: I think the current API does not allow marking as sensitive, it allows only adding more keys to the secret. Maybe a small note somewhere saying `if the schema did not mark it as sensitive and it shows up in CRD, you can make it sensitive by changing the schema`

docs/configuring-a-resource.md
```diff
@@ -0,0 +1,410 @@
...
+  aws_access_key_id: QUtJQVk0QUZUVFNFNDI2TlhKS0I=
+  aws_secret_access_key: ABCxyzRedacted==
+kind: Secret
+```
+
+### Late Initialization Behavior
+Terrajet runtime automatically performs late-initialization during
+an [`external.Observe`] call with means of runtime reflection.
+State of the world observed by Terraform CLI is used to initialize
+any `nil`-valued pointer parameters in the managed resource's `spec`.
```
muvaf, line 300: I think this is too technical of a description late-init. Maybe saying something along the lines of `it will fill the blanks in spec by defaults from the cloud provider` would be easier to understand what it's for.

docs/configuring-a-resource.md
```diff
@@ -0,0 +1,410 @@
...
+func Configure(p *config.Provider) {
+	p.AddResourceConfigurator("azurerm_subnet", func(r *config.Resource) {
+		r.LateInitializer = config.LateInitializer{
+			IgnoredFields: []string{"address_prefix"},
+		}
+	})
+}
+```
+
+In most cases, custom late-initialization configuration will not be necessary.
```
muvaf, line 363: Up until this point users still don't know why they might need to customize this behavior. So, I think having an example case at the beginning like `if you see that behavior, then you need to customize this by following these steps`.

docs/generating-a-provider.md
```diff
@@ -0,0 +1,391 @@
+# Generating a Crossplane Provider
```
muvaf, line 1: We should link this from `README.md`

docs/generating-a-provider.md
```diff
@@ -0,0 +1,391 @@
...
+
+   To limit the resources to be generated, we need to provide an include list
+   option with `tjconfig.WithIncludeList` in file `config/provider.go`:
+
+   ```go
+   pc := tjconfig.NewProvider(resourceMap, resourcePrefix, modulePath,
+       tjconfig.WithDefaultResourceFn(defaultResourceFn),
+       tjconfig.WithIncludeList([]string{
+           "github_repository$",
+           "github_branch$",
```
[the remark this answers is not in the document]
displague: Including a commented out `.*` example would demonstrate that regex is available and showcase (what I imagine) would be the most common setting.

docs/generating-a-provider.md
displague: I was about to leave the same comment.

I'm about to run through this guide for a second provider and I plan on skipping steps 4, 6, and 7.  I'll use `.*`.

Once I have all of the CRDs generated I will revisit step 7 to set up references. I'll revisit step 6 if I have to remove any resources that don't convert well.

docs/generating-a-provider.md
```diff
@@ -0,0 +1,391 @@
...
+   mkdir config/branch
+   ```
+
+   ```bash
+   cat <<EOF > config/repository/config.go
+   package repository
+
+   import "github.com/crossplane-contrib/terrajet/pkg/config"
+
+   func Customize(p *config.Provider) {
```
displague, line 166: Let's not make `reviewable` angry 😄 :

```
exported function `Customize` should have comment or be unexported (golint)
func Customize(p *config.Provider) {
```

docs/generating-a-provider.md
```diff
@@ -0,0 +1,391 @@
...
+       p.AddResourceConfigurator("github_branch", func(r *config.Resource) {
+   
+           // we need to override the default group that terrajet generated for
+           // this resource, which would be "github" 
+           r.Group = "branch"
+   
+           // Identifier for this resource is assigned by the provider. In other
+           // words it is not simply the name of the resource.
+           r.ExternalName = config.IdentifierFromProvider
```
displague, line 192: Does it seem like this should be the default? Beyond the description you've provided can you include an example of when this might be used or not used?

displague: I see you go into more detail in https://github.com/crossplane-contrib/terrajet/pull/130/files#diff-d1f30f3feb80f8b6f5335229713aa3a7ee61093ffe823517e9204422b255251aR15 -- It would help if this step of the guide was part of the other guide.

docs/generating-a-provider.md
```diff
@@ -0,0 +1,391 @@
...
+   cat <<EOF > config/repository/config.go
+   package repository
+
+   import "github.com/crossplane-contrib/terrajet/pkg/config"
+
+   func Customize(p *config.Provider) {
+       p.AddResourceConfigurator("github_repository", func(r *config.Resource) {
+   
+           // we need to override the default group that terrajet generated for
+           // this resource, which would be "github"
```
displague, line 170: Why would the generated group be `github` rather than the last part of any `_` string?

docs/generating-a-provider.md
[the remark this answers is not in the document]
displague: It is difficult to exclude patterns. Perhaps a WithExcludeList feature would help.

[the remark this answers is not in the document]
displague: nm - I found WithSkipList

docs/generating-a-provider.md
```diff
@@ -0,0 +1,391 @@
...
+        export ProviderNameUpper=GitHub
+        ```
+
+    2. Replace all occurrences of `template` with your provider name:
+
+        ```bash
+        git grep -l 'template' -- './*' ':!build/**' ':!go.sum' | xargs sed -i.bak "s/template/${ProviderNameLower}/g"
+        git grep -l 'Template' -- './*' ':!build/**' ':!go.sum' | xargs sed -i.bak "s/Template/${ProviderNameUpper}/g"
+        # Clean up the .bak files created by sed
+        git clean -fd
```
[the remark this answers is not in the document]
turkenh: thanks, I preferred this since it works same for all OSs

docs/generating-a-provider.md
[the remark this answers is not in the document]
turkenh: @zzxwill this is already available as a [variable in makefile](https://github.com/crossplane-contrib/provider-jet-template/blob/d8baa3a1f2e8e3f90dfc34232a18b319985b033b/Makefile#L7). 
Currently, we are using the latest TF cli version and honestly, I would expect that having to use an old version of a cli with a provider not to be that common that worths mentioning here.

docs/generating-a-provider.md
turkenh: these files don't exist at this point and I found this to be more convenient while following the guide, instead of creating directory, file and copy/pasting file content to an editor.

docs/generating-a-provider.md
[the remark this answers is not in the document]
turkenh: @muvaf not sure I got your point. We are already including single resource when we append `$` to end. Otherwise, it will generate all resources starting with `github_repository` or `github_branch`

docs/generating-a-provider.md
turkenh: I'll keep this as is not to confuse people following the guide here, which is also aligned what we agreed here: https://github.com/crossplane-contrib/provider-jet-gcp/pull/2#discussion_r749302880

docs/generating-a-provider.md
turkenh: I would prefer to create an end-to-end experience here. 
Starting from scratch and having some resources provisioned. I am a bit hesitant that if people try to generate all resources at this point, they might be discouraged due to some unexpected issues (e.g. conflicts due to a missing resource configuration).

I've extended the note after step 8 to mention how all resources could be generated and configuring them could be a next step.

docs/configuring-a-resource.md
turkenh: Opened a ticket for that: https://github.com/crossplane-contrib/terrajet/issues/153

docs/configuring-a-resource.md
turkenh: I would say it is still recent :) We can reword that paragraph in a later revisit.

docs/configuring-a-resource.md
turkenh: I believe it is quite exceptional and doesn't worth adding a note here. To the best of my knowledge, we are not aware of such a case yet, right?

turkenh: Maybe, a dedicated section detailing the possibility and reasoning of overriding terraform schema could be added as part of #153

docs/configuring-a-resource.md
turkenh: @ulucinar what do you think? 
In any case, I would prefer to postpone rewdoring this to #153

docs/configuring-a-resource.md
turkenh: Crosslinking to #153 
//cc @ulucinar

docs/generating-a-provider.md
turkenh: Terrajet is doing best in terms of assigning a group, but it is always easily possible to override them. This is the best default based on our experience so far, and changing it would be a bit tricky as it _might_ break existing providers.

turkenh:
This should be ready for another pass!

turkenh:
@sergenyalcin it would be great if could have another look with the latest changes in Terrajet and template repo 🙏

sergenyalcin:
> @sergenyalcin it would be great if could have another look with the latest changes in Terrajet and template repo 🙏

Hi @turkenh I tested the guide with the latest changes. During testing, I noticed that crd manifests are generated with `tf` apigroup names. However in the examples, `jet` name was used as apigroup.

I observed that the template repository does not consume the latest version of terrajet. @ulucinar opened a PR to consume the latest version of terrajet: https://github.com/crossplane-contrib/provider-jet-template/pull/6

I created a repository from this PR and noticed a few changes are necessary due to the use of the latest version. I fixed the problems by my last commit. The `r.Group` field in the `config/repository/config.go` and `config/branch/config.go` was converted to `r.ShortGroup`.

muvaf (approved):
Thanks @turkenh ! Looks great!

Issue #119, opened 2021-10-18
Allow external-name annotation to be constructed from spec and spec fields to be initialized from external-name


ulucinar:
### What problem are you facing?

Like manually coded in native providers, in certain cases, we need a mechanism in terrajet to allow the external-name to be 
constructed from information available from spec & provider configuration. This should also work in the opposite direction, i.e., when an external name is specified by the user, we should be able to late-initialize certain spec fields from the information made available in the external-name. 

### How could Terrajet help solve your problem?

An example from tf-provider-azure is the `v1alpha1.ResourceGroup`:

```
apiVersion: resource.azure.tf.crossplane.io/v1alpha1
kind: ResourceGroup
metadata:
  name: example
  annotations:
    crossplane.io/external-name: /subscriptions/<subscription id>/resourceGroups/example
spec:
  forProvider:
    name: example-resourcegroup
    location: "East US"
    tags:
      provisioner: crossplane
  providerConfigRef:
    name: example
```
In this example, ideally, the users should be able to:
- Specify the required parameters for a resource group, i.e. name & location in the spec to provision a new resource group, and the provider correctly populates the external-name annotation (which is the current case btw)
- Specify only the external-name annotation in the manifest, and the provider should be able to populate the resource group name via a coded extension point (probably part of the `Terraformed` interface).

Issue #104, opened 2021-10-11
Verify our assumption of resources always imported with id


turkenh:
### What problem are you facing?

So far, we had the assumption of using the `id` in terraform attributes to import the resource.  However, it looks like this assumption does not hold always. Noticed that `aws_ecs_cluster` needs to be imported with `name`, however, `id` keeps `arn` as per [documentation](https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/ecs_cluster#import).

### How could Terrajet help solve your problem?

We need to test provision/import `aws_ecs_cluster` and make sure we could import it

PR #132, opened 2021-11-02
Ability to override ID operations

muvaf:

Allows users to override the default `TF State ID -> External Name` and `External Name -> TF State ID` flows. Note that we already have `SetIdentifierArgumentFn` endpoint but that's for overriding `External Name -> Main TF File`. Details about why we're doing this are in the issue.

Depends on https://github.com/crossplane-contrib/terrajet/pull/128 . See the last 3 commits for actual diff.

Fixes https://github.com/crossplane-contrib/terrajet/issues/119
Fixes https://github.com/crossplane-contrib/terrajet/issues/104

I have:

Unit tests and see https://github.com/crossplane-contrib/provider-tf-azure/pull/84

### DIFF
--- a/pkg/config/defaults.go
+++ b/pkg/config/defaults.go
@@ -32,7 +32,7 @@ var (
 			base["name"] = name
 		},
 		GetNameFn: IDAsName,
-		SetIDFn:   NameAsID,
+		GetIDFn:   NameAsID,
 		OmittedFields: []string{
 			"name",
 			"name_prefix",
@@ -45,7 +45,7 @@ var (
 	IdentifierFromProvider = ExternalName{
 		SetIdentifierArgumentFn: NopSetIdentifierArgument,
 		GetNameFn:               IDAsName,
-		SetIDFn:                 NameAsID,
+		GetIDFn:                 NameAsID,
 		DisableNameInitializer:  true,
 	}
 
--- a/pkg/config/defaults.go
+++ b/pkg/config/defaults.go
@@ -23,7 +23,7 @@ import (
 	"github.com/iancoleman/strcase"
 )
 
-// Common ExternalName configurations.
+// Commonly used resource configurations.
 var (
 	// NameAsIdentifier uses "name" field in the arguments as the identifier of
 	// the resource.
@@ -62,15 +62,16 @@ var (
 
 	// NopSensitive does nothing.
 	NopSensitive = Sensitive{
-		AdditionalConnectionDetailsFn: func(_ map[string]interface{}) (map[string][]byte, error) {
-			return nil, nil
-		},
+		AdditionalConnectionDetailsFn: NopAdditionalConnectionDetails,
 	}
 )
 
+// ResourceOption allows setting optional fields of a Resource object.
+type ResourceOption func(*Resource)
+
 // DefaultResource keeps an initial default configuration for all resources of a
 // provider.
-func DefaultResource(name string, terraformSchema *schema.Resource) *Resource {
+func DefaultResource(name string, terraformSchema *schema.Resource, opts ...ResourceOption) *Resource {
 	words := strings.Split(name, "_")
 	// As group name we default to the second element if resource name
 	// has at least 3 elements, otherwise, we took the first element as
@@ -90,7 +91,7 @@ func DefaultResource(name string, terraformSchema *schema.Resource) *Resource {
 		kind = strcase.ToCamel(words[1])
 	}
 
-	return &Resource{
+	r := &Resource{
 		Name:              name,
 		TerraformResource: terraformSchema,
 		ShortGroup:        group,
@@ -100,4 +101,8 @@ func DefaultResource(name string, terraformSchema *schema.Resource) *Resource {
 		References:        map[string]Reference{},
 		Sensitive:         NopSensitive,
 	}
+	for _, f := range opts {
+		f(r)
+	}
+	return r
 }
--- a/pkg/config/defaults.go
+++ b/pkg/config/defaults.go
@@ -96,7 +96,6 @@ func DefaultResource(name string, terraformSchema *schema.Resource) *Resource {
 		ShortGroup:        group,
 		Kind:              kind,
 		Version:           "v1alpha1",
-		IDFieldName:       "id",
 		ExternalName:      NameAsIdentifier,
 		References:        map[string]Reference{},
 		Sensitive:         NopSensitive,
--- a/pkg/config/defaults.go
+++ b/pkg/config/defaults.go
@@ -31,6 +31,8 @@ var (
 		SetIdentifierArgumentFn: func(base map[string]interface{}, name string) {
 			base["name"] = name
 		},
+		GetNameFn: IDAsName,
+		SetIDFn:   NameAsID,
 		OmittedFields: []string{
 			"name",
 			"name_prefix",
@@ -42,6 +44,8 @@ var (
 	// vpc-2213das instead of letting user choose a name.
 	IdentifierFromProvider = ExternalName{
 		SetIdentifierArgumentFn: NopSetIdentifierArgument,
+		GetNameFn:               IDAsName,
+		SetIDFn:                 NameAsID,
 		DisableNameInitializer:  true,
 	}
 
--- a/pkg/config/defaults.go
+++ b/pkg/config/defaults.go
@@ -31,8 +31,8 @@ var (
 		SetIdentifierArgumentFn: func(base map[string]interface{}, name string) {
 			base["name"] = name
 		},
-		GetNameFn: IDAsName,
-		GetIDFn:   NameAsID,
+		GetExternalNameFn: IDAsExternalName,
+		GetIDFn:           ExternalNameAsID,
 		OmittedFields: []string{
 			"name",
 			"name_prefix",
@@ -44,8 +44,8 @@ var (
 	// vpc-2213das instead of letting user choose a name.
 	IdentifierFromProvider = ExternalName{
 		SetIdentifierArgumentFn: NopSetIdentifierArgument,
-		GetNameFn:               IDAsName,
-		GetIDFn:                 NameAsID,
+		GetExternalNameFn:       IDAsExternalName,
+		GetIDFn:                 ExternalNameAsID,
 		DisableNameInitializer:  true,
 	}
 
--- a/pkg/config/defaults.go
+++ b/pkg/config/defaults.go
@@ -59,6 +59,13 @@ var (
 			"internal/controller/providerconfig",
 		},
 	}
+
+	// NopSensitive does nothing.
+	NopSensitive = Sensitive{
+		AdditionalConnectionDetailsFn: func(_ map[string]interface{}) (map[string][]byte, error) {
+			return nil, nil
+		},
+	}
 )
 
 // DefaultResource keeps an initial default configuration for all resources of a
@@ -86,13 +93,12 @@ func DefaultResource(name string, terraformSchema *schema.Resource) *Resource {
 	return &Resource{
 		Name:              name,
 		TerraformResource: terraformSchema,
-
-		ShortGroup: group,
-		Kind:       kind,
-		Version:    "v1alpha1",
-
-		IDFieldName:  "id",
-		ExternalName: NameAsIdentifier,
-		References:   map[string]Reference{},
+		ShortGroup:        group,
+		Kind:              kind,
+		Version:           "v1alpha1",
+		IDFieldName:       "id",
+		ExternalName:      NameAsIdentifier,
+		References:        map[string]Reference{},
+		Sensitive:         NopSensitive,
 	}
 }
--- a/pkg/config/defaults_test.go
+++ b/pkg/config/defaults_test.go
@@ -19,9 +19,8 @@ package config
 import (
 	"testing"
 
-	"github.com/google/go-cmp/cmp/cmpopts"
-
 	"github.com/google/go-cmp/cmp"
+	"github.com/google/go-cmp/cmp/cmpopts"
 	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
 )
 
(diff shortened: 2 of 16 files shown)


turkenh:
@muvaf would this mean, users no longer reliably use Terraform import id as an `external name`? I was liking that what to use as an `external name` is documented for all resources, especially for import cases.

If we chose to still use Terraform import ids as external names, I believe we could cover all cases with two functions as below:

1. CR (metadata.name + spec) => identifier attributes (e.g. `attributes["id"]` and `attributes["bucket-name"]` set)
2. attributes["id"] -> CR (external-name and spec)

I think this would mean us not marking some (required) fields as not required so that they can be late initialized during import.

muvaf:
Synced with @turkenh about his proposal and we decided not to proceed with it because it'd mean to make external name same as ID which is not acceptable in some cases especially in composition where you may not have some parts of the information stored in the ID, such as subscriptionID in Azure.

@ulucinar @turkenh I added your suggestion about the provider config content and opened https://github.com/crossplane-contrib/provider-tf-azure/pull/84 to show how it all would look like.

pkg/config/defaults.go
```diff
@@ -23,14 +23,16 @@ import (
 	"github.com/iancoleman/strcase"
 )
 
-// Common ExternalName configurations.
+// Commonly used resource configurations.
 var (
 	// NameAsIdentifier uses "name" field in the arguments as the identifier of
 	// the resource.
 	NameAsIdentifier = ExternalName{
 		SetIdentifierArgumentFn: func(base map[string]interface{}, name string) {
```
turkenh, line 31: nit: could we make it look similar to below configs by defining it as a function and using its name here.

pkg/resource/sensitive.go
```diff
@@ -93,23 +87,35 @@ func GetConnectionDetails(attr map[string]interface{}, tr Terraformed, cfg *conf
...
+			conn = map[string][]byte{}
 		}
 		conn[k] = v
 	}
+
+	// NOTE(muvaf): Having ID in the connection details is not really a must but
+	// just a nice-to-have, hence we don't handle cases where the ID is given
+	// in a field other than "id". cfg.ExternalName.GetNameFn wouldn't really
+	// help because we need the raw ID.
+	if id, ok := attr["id"]; conn != nil && ok {
```
turkenh, line 102: Should we consider dropping including `id` to connection secrets? 

I remember we've added this to support cases where `id` field is required as a connection detail, e.g. id in [aws_iam_access_key](https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/iam_access_key#id). But now we have a way to include any information from the tf schema as a conn detail with `AdditionalConnectionDetailsFn`.

pkg/config/resource.go
```diff
@@ -20,40 +20,59 @@ import (
...
 // to fill attributes with information given in external name.
 type ExternalName struct {
 	// SetIdentifierArgumentFn sets the name of the resource in Terraform argument
-	// map.
-	SetIdentifierArgumentFn SetIdentifierArgumentFn
+	// map, otherwise we cannot know which field in HCL we should treat as identifier.
+	SetIdentifierArgumentFn SetIdentifierArgumentsFn
+
+	// GetNameFn returns the external name extracted from TF State. In most cases,
+	// "id" key of TF State should be returned.
```
turkenh, line 70: Could we add some quick examples as comments to clarify the intention of these 3 functions further given it is not straightforward to grasp?

pkg/config/resource.go
```diff
@@ -20,40 +20,59 @@ import (
 	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
 )
 
-// SetIdentifierArgumentFn sets the name of the resource in Terraform attributes map.
-type SetIdentifierArgumentFn func(base map[string]interface{}, name string)
+// SetIdentifierArgumentsFn sets the name of the resource in Terraform attributes map,
+// i.e. Main HCL file.
+type SetIdentifierArgumentsFn func(base map[string]interface{}, name string)
```
turkenh, line 25: This `name` argument here is `external-name` right? Should we change it to `externalName` ?

pkg/config/resource.go
```diff
@@ -20,40 +20,59 @@ import (
...
 
 // NopSetIdentifierArgument does nothing. It's useful for cases where the external
 // name is calculated by provider and doesn't have any effect on spec fields.
-func NopSetIdentifierArgument(_ map[string]interface{}, _ string) {}
+var NopSetIdentifierArgument SetIdentifierArgumentsFn = func(_ map[string]interface{}, _ string) {}
+
+// GetIDFn returns the ID to be used in TF State file, i.e. "id" field in
+// terraform.tfstate.
+type GetIDFn func(name string, parameters map[string]interface{}, providerConfig map[string]interface{}) string
```
turkenh, line 33: Similarly here: https://github.com/crossplane-contrib/terrajet/pull/132/files#r751183015

pkg/config/resource.go
```diff
@@ -20,40 +20,59 @@ import (
...
+
+// NameAsID returns the name to be used as ID in TF State file.
+var NameAsID GetIDFn = func(name string, _ map[string]interface{}, _ map[string]interface{}) string {
+	return name
+}
+
+// GetNameFn returns the external name extracted from the TF State.
+type GetNameFn func(tfstate map[string]interface{}) string
```
turkenh, line 41: I believe returning errors in these functions would be helpful for debugging by making what went wrong more explicit, i.e. instead returning an empty string.

pkg/config/resource.go
```diff
@@ -20,40 +20,59 @@ import (
...
+	// map, otherwise we cannot know which field in HCL we should treat as identifier.
+	SetIdentifierArgumentFn SetIdentifierArgumentsFn
+
+	// GetNameFn returns the external name extracted from TF State. In most cases,
+	// "id" key of TF State should be returned.
+	GetNameFn GetNameFn
+
+	// GetIDFn returns the string that will be used as "id" key in TF state. In
+	// most cases, it should return given external name as is.
+	GetIDFn GetIDFn
```
ulucinar, line 75: nit: The immediate context here is (the enclosing type) the `ExternalName` struct, which is a Crossplane concept. I think it will be more explicit to rename this field as `GetTerraformIDFn`.

pkg/resource/sensitive.go
```diff
@@ -65,25 +68,16 @@ type SecretClient interface {
...
 // Terraform attributes and additions connection details configured.
 func GetConnectionDetails(attr map[string]interface{}, tr Terraformed, cfg *config.Resource) (managed.ConnectionDetails, error) {
-	conn, err := GetSensitiveAttributes(attr, tr.GetConnectionDetailsMapping(), tr.GetTerraformResourceIDField())
+	conn, err := GetSensitiveAttributes(attr, tr.GetConnectionDetailsMapping())
 	if err != nil {
 		return nil, errors.Wrap(err, "cannot get connection details")
 	}
 
-	var custom map[string][]byte
-	// TODO(turkenh): Once we have automatic defaulting, remove this if check.
```
ulucinar, line 74: Is the idea here that we will always have a non-zero `AdditionalConnectionDetailsFn` because of the defaulting? I would still prefer to keep this check so that a nil `AdditionalConnectionDetailsFn` continues to behave as a noop.

pkg/resource/sensitive.go
ulucinar: Looks like we pass Terraform attribute map to `AdditionalConnectionDetailsFn`. `id` might incorporate addition info, such as the subscription id for Azure resources, which may not be available in any other attributes. We should consider this in the discussion.

pkg/terraform/files.go
```diff
@@ -61,10 +61,7 @@ func NewFileProducer(ctx context.Context, client resource.SecretClient, dir stri
 	if err = resource.GetSensitiveParameters(ctx, client, tr, params, tr.GetConnectionDetailsMapping()); err != nil {
 		return nil, errors.Wrap(err, "cannot get sensitive parameters")
 	}
-	// TODO(muvaf): Once we have automatic defaulting, remove this if check.
-	if fp.Config.ExternalName.SetIdentifierArgumentFn != nil {
-		fp.Config.ExternalName.SetIdentifierArgumentFn(params, meta.GetExternalName(tr))
-	}
+	fp.Config.ExternalName.SetIdentifierArgumentFn(params, meta.GetExternalName(tr))
```
ulucinar, line 64: I think this was a common pattern we wanted to switch to after defaulting. Please see the above comment.

pkg/resource/sensitive.go
muvaf: @ulucinar The attribute map we pass to `AdditionalConnectionDetailsFn` is the full TFState, which is where we pull the `id` from in this logic. So it should be available in that map. Maybe we can rename the argument name of that function from `attr` to `tfstate` to make it more clear.

@turkenh I thought about removing it but then ended up publishing it only if connection detail secret has some content so that we don't publish a connection secret for every resource, thinking that `id` of RDS Instance could be helpful for connecting for example but I do expect developers to write an `AdditionalConnectionDetailsFn` for _connectible_ resources in all cases anyway. I'll just remove it.

pkg/config/defaults.go
muvaf: It's not really much meaningful for this function to live without the `OmittedFields` here as opposed the ones we have like `NopSetIdentifierArgument`, which is useful on its own.

pkg/config/resource.go
```diff
@@ -20,40 +20,59 @@ import (
...
-}
-
 // ExternalName contains all information that is necessary for naming operations,
 // such as removal of those fields from spec schema and calling Configure function
 // to fill attributes with information given in external name.
 type ExternalName struct {
 	// SetIdentifierArgumentFn sets the name of the resource in Terraform argument
-	// map.
-	SetIdentifierArgumentFn SetIdentifierArgumentFn
+	// map, otherwise we cannot know which field in HCL we should treat as identifier.
```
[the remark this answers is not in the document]
muvaf: We had a similar discussion with @turkenh and my main reasons for explicitly relying on existence of these functions are mainly:
* If we assume no-op when it's not specified, we're making a decision on behalf of the developers and if it covers enough ground for initial tests, then they'll be aware of that decision later on rather than immediately.
* We have `DefaultResource` that makes this decision in a single place and it covers all those functions and also provide ways to override that provider-wide default so that they can very easily surface how that decision is made instead of diving into code where these functions are called.

So I'd prefer to keep requiring existence of this function. We can possibly have a validation somewhere in code-generation to check if they are `nil` so that we panic earlier.

pkg/resource/sensitive.go
muvaf: See my comment in https://github.com/crossplane-contrib/terrajet/pull/132#discussion_r751677581

pkg/config/resource.go
muvaf: Added it. I also added `ctx` to `GetIDFn` so that we make queries there, like getting account id by making `get-caller-identity` call in AWS to construct the full ARN.

pkg/config/resource.go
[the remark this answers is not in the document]
muvaf: > i.e., a "producer" puts key,value pairs to be consumed by a specific instance of the GetIDFn via known keys.

Could you elaborate? Right now, we can only give generic information to this function like those given, and if users want to add more they can do that in runtime by returning anonymous function that take value from somewhere else.

> Thus we may consider returning an error here and breaking early before running any Terraform commands in FileProducer.WriteTFState.

Thanks, added.

pkg/config/resource.go
```diff
@@ -20,40 +20,59 @@ import (
...
 // name is calculated by provider and doesn't have any effect on spec fields.
-func NopSetIdentifierArgument(_ map[string]interface{}, _ string) {}
+var NopSetIdentifierArgument SetIdentifierArgumentsFn = func(_ map[string]interface{}, _ string) {}
+
+// GetIDFn returns the ID to be used in TF State file, i.e. "id" field in
+// terraform.tfstate.
+type GetIDFn func(name string, parameters map[string]interface{}, providerConfig map[string]interface{}) string
+
+// NameAsID returns the name to be used as ID in TF State file.
+var NameAsID GetIDFn = func(name string, _ map[string]interface{}, _ map[string]interface{}) string {
```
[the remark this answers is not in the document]
muvaf: I added them on purpose actually so that we get compilation errors here right away if they don't adhere to those types. Otherwise, they'd remain as anonymous functions until used somewhere to return compile error.

pkg/config/resource.go
```diff
@@ -20,40 +20,59 @@ import (
...
+	return name
+}
+
+// GetNameFn returns the external name extracted from the TF State.
+type GetNameFn func(tfstate map[string]interface{}) string
+
+// IDAsName returns the TF State ID as external name.
+var IDAsName GetNameFn = func(tfstate map[string]interface{}) string {
+	if id, ok := tfstate["id"].(string); ok {
```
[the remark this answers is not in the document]
muvaf: Added error reporting and made sure we return error if `id` is `""`

pkg/config/resource.go
muvaf: I'd actually prefer to keep it as ID since that's what it's called in the map that this function processes, similarly how `ExternalName` corresponds to `crossplane.io/external-name` annotation.

muvaf:
@ulucinar @turkenh I think I have responded to all comments and made changes, ready for another look 🙂

pkg/config/defaults_test.go
```diff
@@ -0,0 +1,87 @@
...
+*/
+
+package config
+
+import (
+	"testing"
+
+	"github.com/google/go-cmp/cmp/cmpopts"
+
+	"github.com/google/go-cmp/cmp"
```
turkenh, line 24: nit: goimports

turkenh (approved):
LGTM, thanks @muvaf !

PR #137, opened 2021-11-10
Find a unique type name by appending indexes

### LINKED ISSUE #135: Could not generate a unique name for SubnetworkParameters
### What happened?

While trying to generate `google_compute_subnetwork` together with `google_compute_router_nat` resource, Terrajet fails to find a unique name for `SubnetworkParameters`:

```
panic: cannot generate crd for resource google_compute_subnetwork: cannot build types for Subnetwork: cannot build the types: cannot generate parameters type name of Subnetwork: could not generate a unique name for SubnetworkParameters

goroutine 1 [running]:
github.com/crossplane-contrib/terrajet/pkg/pipeline.Run(0xc000636300)
	/Users/hasanturken/Workspace/crossplane-contrib/provider-tf-gcp/.work/pkg/pkg/mod/github.com/crossplane-contrib/terrajet@v0.1.1-0.20211104212137-874bb6ad5cff/pkg/pipeline/run.go:82 +0x2053
main.main()
	/Users/hasanturken/Workspace/crossplane-contrib/provider-tf-gcp/cmd/generator/main.go:10 +0x2a
exit status 2
make[1]: *** [terrajet.run] Error 1
make: *** [generate] Error 2
```

This is because `google_compute_router_nat` resource already have a type named `SubnetworkParameters` generated and  `google_compute_subnetwork` resource coming after it alphabetically (since we are sorting for stability) fails to generate a type for top level parameter struct.

I believe we will need a way to configure type names to handle such cases.

### How can we reproduce it?

Try generating both resources together in provider-tf-gcp.

turkenh:

Fixes #135

I have:

With `provider-tf-gcp` by generating all resources.


pkg/config/provider.go
```diff
@@ -199,7 +199,10 @@ func (p *Provider) SetResourceConfigurator(resource string, c ResourceConfigurat
 // ConfigureResources configures resources with provided ResourceConfigurator's
 func (p *Provider) ConfigureResources() {
 	for name, c := range p.resourceConfigurators {
-		c.Configure(p.Resources[name])
+		// Only call configure if resource included.
+		if _, ok := p.Resources[name]; ok {
```
turkenh, line 203: This is not directly related to this PR but a simple fix to an issue that prevents skipping resources that have custom configurations.

pkg/pipeline/crd.go
```diff
@@ -79,12 +83,30 @@ func (cg *CRDGenerator) Generate(cfg *config.Resource) error {
 	if err != nil {
 		return errors.Wrap(err, "cannot print the type list")
 	}
+	parametersType := ""
+	observationType := ""
+	for _, t := range typeList {
+		if reParametersType.MatchString(strings.TrimPrefix(t.Obj().Name(), cfg.Kind)) {
```
muvaf, line 89: `buildResource` already [returns](https://github.com/crossplane-contrib/terrajet/blob/6d7fb7e/pkg/types/builder.go#L57) the parameter and observation types but we just ignore them at the highest level call. Is there a way to bring that information here instead of doing a search in the list?

pkg/types/builder.go
```diff
@@ -293,6 +293,14 @@ func (g *Builder) generateTypeName(suffix string, names ...string) (string, erro
 	if g.Package.Scope().Lookup(n) == nil {
 		return n, nil
 	}
+	// start from 2 considering the 1st of this type is the one without an
+	// index.
+	for i := 2; i < 10; i++ {
+		nn := fmt.Sprintf("%s%d", n, i)
```
muvaf, line 299:
```suggestion
		nn := fmt.Sprintf("%s_%d", n, i)
```
nitpick

pkg/types/builder.go
```diff
@@ -293,6 +293,14 @@ func (g *Builder) generateTypeName(suffix string, names ...string) (string, erro
 	if g.Package.Scope().Lookup(n) == nil {
 		return n, nil
 	}
+	// start from 2 considering the 1st of this type is the one without an
+	// index.
+	for i := 2; i < 10; i++ {
```
muvaf, line 298: I wonder if we can get away with appending the CRD kind name repeatedly. Instead of `SubnetworkParameters2`, how would `SubnetworkParameters_Subnetwork` or `Subnetwork_SubnetworkParameters` look? Feels like because they are in the same package, it's kind of too easy to confuse the one used by one resource with the other when it's only a number that differentiates them.

turkenh: Given this is just generated code and we don't expect anyone to deal with modifying those types, I prefer numbers.
One case could be people importing these types to interact programmatically but still, in that case, I believe using the wrong type would cause compilation errors hence easy to figure out.

Even if we go with appending the kind name as you suggested, we will still have both `SubnetworkParameters` and `SubnetworkParameters_Subnetwork` in the same package and the former would belong to `RouterNat` resource instead of `Subnetwork` which is still confusing.

muvaf: I was thinking about someone importing this package and using the types rather than modifying. In such cases, the types being generated or not doesn't really matter for them and Go types having numbers in the name doesn't look very idiomatic. I'm OK with both approaches though.

pkg/pipeline/crd.go
turkenh: Great idea, done!

turkenh:
@muvaf your comments got resolved!

pkg/types/builder.go
```diff
@@ -39,28 +39,38 @@ const (
...
+
+	ForProviderType *types.Named
+	AtProviderType  *types.Named
+}
+
 // Builder is used to generate Go type equivalence of given Terraform schema.
 type Builder struct {
 	Package *types.Package
-
-	genTypes []*types.Named
```
muvaf, line 51: Originally, I'd have preferred to not have them as properties at all since they are very specific to a single call of `Build` but I had to do it so that recursive calls can access it without having to pass them around all the time. What do you think about keeping these private and return a `Generated` instance that is created and filled up in `Build` as the result?

pkg/types/builder.go
```diff
@@ -39,28 +39,38 @@ const (
...
+func (g *Builder) Build(cfg *config.Resource) (Generated, error) {
+	var err error
+	g.ForProviderType, g.AtProviderType, err = g.buildResource(cfg.TerraformResource, cfg, nil, nil, cfg.Kind)
+	return g.Generated, errors.Wrapf(err, "cannot build the Types")
 }
 
 func (g *Builder) buildResource(res *schema.Resource, cfg *config.Resource, tfPath []string, xpPath []string, names ...string) (*types.Named, *types.Named, error) { //nolint:gocyclo
 	// NOTE(muvaf): There can be fields in the same CRD with same name but in
-	// different types. Since we generate the type using the field name, there
+	// different Types. Since we generate the type using the field name, there
```
muvaf, line 73: Seems like there are quite a few unintended changes from `types` to `Types` done by IDE.

pkg/types/builder.go
```diff
@@ -39,28 +39,38 @@ const (
...
-		comments: twtypes.Comments{},
+		Package: pkg,
+		Generated: Generated{
+			Comments: twtypes.Comments{},
+		},
 	}
 }
 
+// Generated stores the information of the generated types
+type Generated struct {
```
muvaf, line 50: I don't have a great suggestion but `Generated`/`types.Generated` feels too generic.

`types.TypeTree`
```suggestion
// TypeTree contains all the types and their sibling relations.
type TypeTree struct {
```

`types.CustomResourceDefinition`
```suggestion
// CustomResourceDefinition includes all Go types necessary to print a CustomResourceDefinition Go struct.
type CustomResourceDefinition struct {
```

`types.Managed`
```suggestion
// Managed includes all Go types necessary to print the Go structs for a Managed resource.
type Managed struct {
```

turkenh: I cannot claim that `types.Generated` is great but I couldn't find a better alternative and honestly would prefer it to the alternatives you suggested above :)

I believe it is not that bad and actually conveys its purpose: `types.Generated` is a struct that holds _Generated_ _types_

muvaf (approved):
LGTM! It'd be great if you can add tests for `generateTypeName`.

turkenh:
> LGTM! It'd be great if you can add tests for `generateTypeName`.

@muvaf indeed I attempted to add but postponed to a larger refactor since I wanted to change the function signature as follows for easier testability:

```diff
- func (g *Builder) generateTypeName(suffix string, names ...string) (string, error) {

+ type lookupFunc func(name string) types.Object
+ func generateTypeName(lookup lookupFunc, suffix string, names ...string) (string, error) {
```

If this sounds ok to you, I can add unit tests with the above changes.

muvaf:
@turkenh you can actually mock the package and its scope like the following without changing the signature: https://github.com/crossplane/crossplane-tools/blob/fa70b84/internal/method/resolver_test.go#L229

turkenh:
@muvaf added unit tests without changing the function signature.

PR #138, opened 2021-11-11
Do not run configurator chain on a not-included resource

ulucinar:

If we have a configurator defined on a not-included resource, we currently observe a panic. This PR suggests a change that skips running the configurator chain on a not-included resource, even if a configurator has been specified for it. This allows us to keep/maintain configuration code independently of the set of resources included/excluded.

I have:

Tested with a `provider-tf-azure` [PR](https://github.com/crossplane-contrib/provider-tf-azure/pull/83).


turkenh (approved):
Thanks @ulucinar!
I hit this as well and indeed [covered](https://github.com/crossplane-contrib/terrajet/pull/137/files#r746590889) in my open PR, however, we can just merge this without waiting that one.

Issue #99, opened 2021-10-08
Logs and events may contain provider credentials


turkenh:
### What happened?

Trying to create an elasticache cluster and it failed for some reason I am still debugging, but realized another problem which is terraform logs and event on cr contains `access_key/secret_key`:

See the following (intentionally replaced with `REDACTED`): 

```
  conditions:
  - lastTransitionTime: "2021-10-08T17:02:05Z"
    message: |-
      observe failed: cannot run refresh: cannot refresh: {"@level":"info","@message":"Terraform 1.0.4","@module":"terraform.ui","@timestamp":"2021-10-08T20:02:04.398569+03:00","terraform":"1.0.4","type":"version","ui":"0.1.0"}
      {"@level":"info","@message":"\nInterrupt received.\nPlease wait for Terraform to exit or data loss may occur.\nGracefully shutting down...\n","@module":"terraform.ui","@timestamp":"2021-10-08T20:02:05.046946+03:00","type":"log"}
[...]

PR #143, opened 2021-11-15
Extract Terraform error messages from JSON logs

### LINKED ISSUE #44: Proper error reporting and eventing

I am trying to create a VPC with `provider-tf-aws` but it is not getting ready, when I describe it, I see the following output:

```
Name:         sample-vpc
Namespace:
Labels:       <none>
Annotations:  <none>
API Version:  vpc.aws.tf.crossplane.io/v1alpha1
Kind:         Vpc
Metadata:
  Creation Timestamp:  2021-09-01T09:41:19Z
  Finalizers:
    finalizer.managedresource.crossplane.io
  Generation:  1
  # omit ManagedFields
  Resource Version:  2817
  UID:               578e6a58-636b-4e2f-82bd-5bf1c691e9fc
Spec:
  Deletion Policy:  Delete
  For Provider:
    Cidr Block:  10.0.0.0/16
    Region:      us-east-2
    Tags:
      Name:  DemoVpc
    Tags All:
      Name:  DemoVpc
  Provider Config Ref:
    Name:  example
Status:
  At Provider:
    Arn:
    Default Network Acl Id:
    Default Route Table Id:
    Default Security Group Id:
    Dhcp Options Id:
    ipv6AssociationId:
    ipv6CidrBlock:
    Main Route Table Id:
    Owner Id:
  Conditions:
    Last Transition Time:  2021-09-01T09:41:25Z
    Reason:                Creating
    Status:                False
    Type:                  Ready
    Last Transition Time:  2021-09-01T09:42:54Z
    Reason:                ReconcileSuccess
    Status:                True
    Type:                  Synced
Events:
  Type     Reason                        Age                From                                  Message
  ----     ------                        ----               ----                                  -------
  Normal   CreatedExternalResource       6s (x10 over 95s)  managed/vpc.vpc.aws.tf.crossplane.io  Successfully requested creation of external resource
  Warning  CannotCreateExternalResource  6s (x3 over 82s)   managed/vpc.vpc.aws.tf.crossplane.io  failed to update: cannot update with tf cli: failed to apply Terraform configuration:
Terraform used the selected providers to generate the following execution
plan. Resource actions are indicated with the following symbols:
  + create

Terraform will perform the following actions:

  # aws_vpc.sample-vpc will be created
  + resource "aws_vpc" "sample-vpc" {
      + arn                              = (known after apply)
      + assign_generated_ipv6_cidr_block = false
      + cidr_block                       = "10.0.0.0/16"
      + default_network_acl_id           = (known after apply)
      + default_route_table_id           = (known after apply)
      + default_security_group_id        = (known after apply)
      + dhcp_options_id                  = (known after apply)
      + enable_classiclink               = (known after apply)
      + enable_classiclink_dns_support   = (known after apply)
      + enable_dns_hostnames             = (known after apply)
      + enable_dns_support               = true
      + id                               = (known after apply)
      + instance_tenancy                 = "default"
      + ipv6_association_id              = (known after apply)
      + ipv6_cidr_block                  = (known after apply)
      + main_route_table_id              = (known after apply)
      + owner_id                         = (known after apply)
      + tags                             = {
          + "Name" = "DemoVpc"
        }
      + tags_all                         = {
          + "Name" = "DemoVpc"
        }
    }

Plan: 1 to add, 0 to change, 0 to destroy.
aws_vpc.sample-vpc: Creating...
```

However, when I check the controller logs, I can see the actual problem somewhere in a long log message:

```
2021-09-01T12:43:01.760+0300    INFO    provider-tf-aws Failed to run Terraform CLI     {"tfcli-version": "0.0.0", "args": ["apply", "-auto-approve", "-input=false"], "executable": "terraform", "cwd": "/var/folders/jb/zwwlz42935308fcydsp4h05h0000gn/T/ws-93c57199f76ed373701483fb75b20d8d22840904b04ce4f7a1093882131d1b99", "stderr": "\u001b[31m╷\u001b[0m\u001b[0m\n\u001b[31m│\u001b[0m \u001b[0m\u001b[1m\u001b[31mError: \u001b[0m\u001b[0m\u001b[1mError creating VPC: VpcLimitExceeded: The maximum number of VPCs has been reached.\n\u001b[31m│\u001b[0m \u001b[0m\tstatus code: 400, request id: e26f86cf-8bb1-40f3-a5f8-6c9e36351b52\u001b[0m\n\u001b[31m│\u001b[0m \u001b[0m\n\u001b[31m│\u001b[0m \u001b[0m\u001b[0m  with aws_vpc.sample-vpc,\n\u001b[31m│\u001b[0m \u001b[0m  on main.tf.json line 20, in resource.aws_vpc.sample-vpc.tags_all:\n\u001b[31m│\u001b[0m \u001b[0m  20:             \"sample-vpc\": {\"assign_generated_ipv6_cidr_block\":null,\"cidr_block\":\"10.0.0.0/16\",\"enable_classiclink\":null,\"enable_classiclink_dns_support\":null,\"enable_dns_hostnames\":null,\"enable_dns_support\":null,\"instance_tenancy\":null,\"tags\":{\"Name\":\"DemoVpc\"},\"tags_all\":{\"Name\":\"DemoVpc\"}\u001b[4m}\u001b[0m\u001b[0m\n\u001b[31m│\u001b[0m \u001b[0m\n\u001b[31m╵\u001b[0m\u001b[0m\n", "stdout": "\nTerraform used the selected providers to generate the following execution\nplan. Resource actions are indicated with the following symbols:\n  \u001b[32m+\u001b[0m create\n\u001b[0m\nTerraform will perform the following actions:\n\n\u001b[1m  # aws_vpc.sample-vpc\u001b[0m will be created\u001b[0m\u001b[0m\n\u001b[0m  \u001b[32m+\u001b[0m\u001b[0m resource \"aws_vpc\" \"sample-vpc\" {\n      \u001b[32m+\u001b[0m \u001b[0m\u001b[1m\u001b[0marn\u001b[0m\u001b[0m                              = (known after apply)\n      \u001b[32m+\u001b[0m \u001b[0m\u001b[1m\u001b[0massign_generated_ipv6_cidr_block\u001b[0m\u001b[0m = false\n      \u001b[32m+\u001b[0m \u001b[0m\u001b[1m\u001b[0mcidr_block\u001b[0m\u001b[0m                       = \"10.0.0.0/16\"\n      \u001b[32m+\u001b[0m \u001b[0m\u001b[1m\u001b[0mdefault_network_acl_id\u001b[0m\u001b[0m           = (known after apply)\n      \u001b[32m+\u001b[0m \u001b[0m\u001b[1m\u001b[0mdefault_route_table_id\u001b[0m\u001b[0m           = (known after apply)\n      \u001b[32m+\u001b[0m \u001b[0m\u001b[1m\u001b[0mdefault_security_group_id\u001b[0m\u001b[0m        = (known after apply)\n      \u001b[32m+\u001b[0m \u001b[0m\u001b[1m\u001b[0mdhcp_options_id\u001b[0m\u001b[0m                  = (known after apply)\n      \u001b[32m+\u001b[0m \u001b[0m\u001b[1m\u001b[0menable_classiclink\u001b[0m\u001b[0m               = (known after apply)\n      \u001b[32m+\u001b[0m \u001b[0m\u001b[1m\u001b[0menable_classiclink_dns_support\u001b[0m\u001b[0m   = (known after apply)\n      \u001b[32m+\u001b[0m \u001b[0m\u001b[1m\u001b[0menable_dns_hostnames\u001b[0m\u001b[0m             = (known after apply)\n      \u001b[32m+\u001b[0m \u001b[0m\u001b[1m\u001b[0menable_dns_support\u001b[0m\u001b[0m               = true\n      \u001b[32m+\u001b[0m \u001b[0m\u001b[1m\u001b[0mid\u001b[0m\u001b[0m                               = (known after apply)\n      \u001b[32m+\u001b[0m \u001b[0m\u001b[1m\u001b[0minstance_tenancy\u001b[0m\u001b[0m                 = \"default\"\n      \u001b[32m+\u001b[0m \u001b[0m\u001b[1m\u001b[0mipv6_association_id\u001b[0m\u001b[0m              = (known after apply)\n      \u001b[32m+\u001b[0m \u001b[0m\u001b[1m\u001b[0mipv6_cidr_block\u001b[0m\u001b[0m                  = (known after apply)\n      \u001b[32m+\u001b[0m \u001b[0m\u001b[1m\u001b[0mmain_route_table_id\u001b[0m\u001b[0m              = (known after apply)\n      \u001b[32m+\u001b[0m \u001b[0m\u001b[1m\u001b[0mowner_id\u001b[0m\u001b[0m                         = (known after apply)\n      \u001b[32m+\u001b[0m \u001b[0m\u001b[1m\u001b[0mtags\u001b[0m\u001b[0m                             = {\n          \u001b[32m+\u001b[0m \u001b[0m\"Name\" = \"DemoVpc\"\n        }\n      \u001b[32m+\u001b[0m \u001b[0m\u001b[1m\u001b[0mtags_all\u001b[0m\u001b[0m                         = {\n          \u001b[32m+\u001b[0m \u001b[0m\"Name\" = \"DemoVpc\"\n        }\n    }\n\n\u001b[0m\u001b[1mPlan:\u001b[0m 1 to add, 0 to change, 0 to destroy.\n\u001b[0m\u001b[0m\u001b[1maws_vpc.sample-vpc: Creating...\u001b[0m\u001b[0m\n", "error": "exit status 1"}
```

Which is indeed: 
```
Error creating VPC: VpcLimitExceeded: The maximum number of VPCs has been reached.
```

We need to ensure to return actual errors in the form of events on the managed resource. 
Also, similar to other crossplane providers, we should hide (per resource) logs at info level.

ulucinar:

Fixes #44

This PR proposes changes that parse the actual error messages from respective JSON-formatted Terraform CLI logs in case the Terraform CLI pipeline fails.

I have:

On top of https://github.com/crossplane-contrib/provider-tf-azure/pull/83, an errored condition without these changes results in:
```
status:
  atProvider: {}
  conditions:
  - lastTransitionTime: "2021-11-15T14:05:25Z"
    reason: Creating
    status: "False"
    type: Ready
  - lastTransitionTime: "2021-11-15T14:05:25Z"
    message: |-
      create failed: cannot apply: cannot apply: {"@level":"info","@message":"Terraform 1.0.3","@module":"terraform.ui","@timestamp":"2021-11-15T17:05:05.983000+03:00","terraform":"1.0.3","type":"version","ui":"0.1.0"}
      {"@level":"info","@message":"azurerm_resource_group.example-alper-2: Plan to create","@module":"terraform.ui","@timestamp":"2021-11-15T17:05:15.213746+03:00","change":{"resource":{"addr":"azurerm_resource_group.example-alper-2","module":"","resource":"azurerm_resource_group.example-alper-2","implied_provider":"azurerm","resource_type":"azurerm_resource_group","resource_name":"example-alper-2","resource_key":null},"action":"create"},"type":"planned_change"}
      {"@level":"info","@message":"Plan: 1 to add, 0 to change, 0 to destroy.","@module":"terraform.ui","@timestamp":"2021-11-15T17:05:15.213823+03:00","changes":{"add":1,"change":0,"remove":0,"operation":"plan"},"type":"change_summary"}
      {"@level":"info","@message":"azurerm_resource_group.example-alper-2: Creating...","@module":"terraform.ui","@timestamp":"2021-11-15T17:05:24.376958+03:00","hook":{"resource":{"addr":"azurerm_resource_group.example-alper-2","module":"","resource":"azurerm_resource_group.example-alper-2","implied_provider":"azurerm","resource_type":"azurerm_resource_group","resource_name":"example-alper-2","resource_key":null},"action":"create"},"type":"apply_start"}
      {"@level":"info","@message":"azurerm_resource_group.example-alper-2: Creation errored after 1s","@module":"terraform.ui","@timestamp":"2021-11-15T17:05:24.985199+03:00","hook":{"resource":{"addr":"azurerm_resource_group.example-alper-2","module":"","resource":"azurerm_resource_group.example-alper-2","implied_provider":"azurerm","resource_type":"azurerm_resource_group","resource_name":"example-alper-2","resource_key":null},"action":"create","elapsed_seconds":1},"type":"apply_errored"}
      {"@level":"error","@message":"Error: A resource with the ID \"/subscriptions/.../resourceGroups/example-resourcegroup-alper\" already exists - to be managed via Terraform this resource needs to be imported into the State. Please see the resource documentation for \"azurerm_resource_group\" for more information.","@module":"terraform.ui","@timestamp":"2021-11-15T17:05:25.033488+03:00","diagnostic":{"severity":"error","summary":"A resource with the ID \"/subscriptions/.../resourceGroups/example-resourcegroup-alper\" already exists - to be managed via Terraform this resource needs to be imported into the State. Please see the resource documentation for \"azurerm_resource_group\" for more information.","detail":"","address":"azurerm_resource_group.example-alper-2","range":{"filename":"main.tf.json","start":{"line":1,"column":228,"byte":227},"end":{"line":1,"column":229,"byte":228}},"snippet":{"context":"resource.azurerm_resource_group.example-alper-2.tags","code":"{\"provider\":{\"azurerm\":{\"features\":{}}},\"resource\":{\"azurerm_resource_group\":{\"example-alper-2\":{\"lifecycle\":{\"prevent_destroy\":true},\"location\":\"East US\",\"name\":\"example-resourcegroup-alper\",\"tags\":{\"provisioner\":\"crossplane\"}}}},\"terraform\":{\"required_providers\":{\"azurerm\":{\"source\":\"hashicorp/azurerm\",\"version\":\"2.78.0\"}}}}","start_line":1,"highlight_start_offset":227,"highlight_end_offset":228,"values":[]}},"type":"diagnostic"}
      : exit status 1
    reason: ReconcileError
    status: "False"
    type: Synced
```

With the proposed changes:
```
status:
  atProvider: {}
  conditions:
  - lastTransitionTime: "2021-11-15T14:05:25Z"
    reason: Creating
    status: "False"
    type: Ready
  - lastTransitionTime: "2021-11-15T14:22:13Z"
    message: 'create failed: cannot apply: apply failed: A resource with the ID "/subscriptions/.../resourceGroups/example-resourcegroup-alper"
      already exists - to be managed via Terraform this resource needs to be imported
      into the State. Please see the resource documentation for "azurerm_resource_group"
      for more information.: '
    reason: ReconcileError
    status: "False"
    type: Synced
```

### DIFF
--- a/pkg/terraform/errors/errors.go
+++ b/pkg/terraform/errors/errors.go
@@ -16,40 +16,156 @@ limitations under the License.
 
 package errors
 
+import (
+	"fmt"
+	"strings"
+
+	jsoniter "github.com/json-iterator/go"
+	"github.com/pkg/errors"
+)
+
+const (
+	levelError = "error"
+)
+
+type tfError struct {
+	logs    []byte
+	tfLogs  []*TerraformLog
+	cause   error
+	message string
+}
+
 type applyFailed struct {
-	log string
+	*tfError
 }
 
-func (a *applyFailed) Error() string {
-	return a.log
+// TerraformLog represents relevant fields of a Terraform CLI JSON-formatted log line
+type TerraformLog struct {
+	Level      string        `json:"@level"`
+	Message    string        `json:"@message"`
+	Diagnostic LogDiagnostic `json:"diagnostic"`
 }
 
-// NewApplyFailed returns a new apply failure error with given logs.
-func NewApplyFailed(log string) error {
-	return &applyFailed{log: log}
+// LogDiagnostic represents relevant fields of a Terraform CLI JSON-formatted
+// log line diagnostic info
+type LogDiagnostic struct {
+	Severity string `json:"severity"`
+	Summary  string `json:"summary"`
+	Detail   string `json:"detail"`
+}
+
+func (t *tfError) Error() string {
+	if t == nil {
+		return ""
+	}
+
+	messages := make([]string, 0, len(t.tfLogs))
+	for _, l := range t.tfLogs {
+		// only use error logs
+		if l == nil || l.Level != levelError {
+			continue
+		}
+		m := l.Message
+		if l.Diagnostic.Severity == levelError && l.Diagnostic.Summary != "" {
+			m = fmt.Sprintf("%s: %s", l.Diagnostic.Summary, l.Diagnostic.Detail)
+		}
+		messages = append(messages, m)
+	}
+	return fmt.Sprintf("%s: %s", t.message, strings.Join(messages, "\n"))
+}
+
+func (t *tfError) Unwrap() error {
+	if t == nil {
+		return nil
+	}
+	return t.cause
+}
+
+func newTFError(message string, cause error, logs []byte) (string, *tfError) {
+	tfError := &tfError{
+		logs:    logs,
+		cause:   cause,
+		message: message,
+	}
+
+	tfLogs, err := parseTerraformLogs(logs)
+	if err != nil {
+		return err.Error(), tfError
+	}
+	tfError.tfLogs = tfLogs
+	return "", tfError
+}
+
+func parseTerraformLogs(logs []byte) ([]*TerraformLog, error) {
+	logLines := strings.Split(string(logs), "\n")
+	tfLogs := make([]*TerraformLog, 0, len(logLines))
+	for _, l := range logLines {
+		log := &TerraformLog{}
+		l := strings.TrimSpace(l)
+		if l == "" {
+			continue
+		}
+		if err := jsoniter.ConfigCompatibleWithStandardLibrary.UnmarshalFromString(l, log); err != nil {
+			return nil, err
+		}
+		tfLogs = append(tfLogs, log)
+	}
+	return tfLogs, nil
+}
+
+// WrapTFError returns a new Terraform CLI failure error with given logs.
+func WrapTFError(message string, cause error, logs []byte) error {
+	if cause == nil {
+		return nil
+	}
+
+	parseError, tfError := newTFError(message, cause, logs)
+	if parseError == "" {
+		return tfError
+	}
+	return errors.WithMessage(tfError, parseError)
+}
+
+// WrapApplyFailed returns a new apply failure error with given logs.
+func WrapApplyFailed(cause error, logs []byte) error {
+	if cause == nil {
+		return nil
+	}
+
+	parseError, tfError := newTFError("apply failed", cause, logs)
+	result := &applyFailed{tfError: tfError}
+	if parseError == "" {
+		return result
+	}
+	return errors.WithMessage(result, parseError)
 }
 
 // IsApplyFailed returns whether error is due to failure of an apply operation.
 func IsApplyFailed(err error) bool {
-	_, ok := err.(*applyFailed)
-	return ok
+	r := &applyFailed{}
+	return errors.As(err, &r)
 }
 
 type destroyFailed struct {
-	log string
+	*tfError
 }
 
-func (a *destroyFailed) Error() string {
-	return a.log
-}
+// WrapDestroyFailed returns a new destroy failure error with given logs.
+func WrapDestroyFailed(cause error, logs []byte) error {
+	if cause == nil {
+		return nil
+	}
 
-// NewDestroyFailed returns a new destroy failure error with given logs.
-func NewDestroyFailed(log string) error {
-	return &destroyFailed{log: log}
+	parseError, tfError := newTFError("destroy failed", cause, logs)
+	result := &destroyFailed{tfError: tfError}
+	if parseError == "" {
+		return result
+	}
+	return errors.WithMessage(result, parseError)
 }
 
 // IsDestroyFailed returns whether error is due to failure of a destroy operation.
 func IsDestroyFailed(err error) bool {
-	_, ok := err.(*destroyFailed)
-	return ok
+	r := &destroyFailed{}
+	return errors.As(err, &r)
 }
--- a/pkg/controller/api_test.go
+++ b/pkg/controller/api_test.go
@@ -20,13 +20,14 @@ import (
 	"context"
 	"testing"
 
-	xpresource "github.com/crossplane/crossplane-runtime/pkg/resource"
-	xpfake "github.com/crossplane/crossplane-runtime/pkg/resource/fake"
-	"github.com/crossplane/crossplane-runtime/pkg/test"
 	"github.com/google/go-cmp/cmp"
 	"github.com/pkg/errors"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	ctrl "sigs.k8s.io/controller-runtime/pkg/manager"
+
+	xpresource "github.com/crossplane/crossplane-runtime/pkg/resource"
+	xpfake "github.com/crossplane/crossplane-runtime/pkg/resource/fake"
+	"github.com/crossplane/crossplane-runtime/pkg/test"
 
 	"github.com/crossplane-contrib/terrajet/pkg/resource"
 	"github.com/crossplane-contrib/terrajet/pkg/resource/fake"
@@ -56,7 +57,7 @@ func TestAPICallbacks_Apply(t *testing.T) {
 						MockGet: test.NewMockGetFn(nil),
 						MockStatusUpdate: func(_ context.Context, obj client.Object, _ ...client.UpdateOption) error {
 							got := obj.(resource.Terraformed).GetCondition(resource.TypeAsyncOperation)
-							if diff := cmp.Diff(resource.AsyncOperationCondition(tjerrors.NewApplyFailed(errBoom.Error())), got); diff != "" {
+							if diff := cmp.Diff(resource.AsyncOperationCondition(tjerrors.WrapApplyFailed(errBoom, nil)), got); diff != "" {
 								t.Errorf("\nApply(...): -want error, +got error:\n%s", diff)
 							}
 							return nil
@@ -64,7 +65,7 @@ func TestAPICallbacks_Apply(t *testing.T) {
 					},
 					Scheme: xpfake.SchemeWith(&fake.Terraformed{}),
 				},
-				err: tjerrors.NewApplyFailed(errBoom.Error()),
+				err: tjerrors.WrapApplyFailed(errBoom, nil),
 			},
 		},
 		"ApplyOperationSucceeded": {
@@ -138,7 +139,7 @@ func TestAPICallbacks_Destroy(t *testing.T) {
 						MockGet: test.NewMockGetFn(nil),
 						MockStatusUpdate: func(_ context.Context, obj client.Object, _ ...client.UpdateOption) error {
 							got := obj.(resource.Terraformed).GetCondition(resource.TypeAsyncOperation)
-							if diff := cmp.Diff(resource.AsyncOperationCondition(tjerrors.NewDestroyFailed(errBoom.Error())), got); diff != "" {
+							if diff := cmp.Diff(resource.AsyncOperationCondition(tjerrors.WrapDestroyFailed(errBoom, nil)), got); diff != "" {
 								t.Errorf("\nApply(...): -want error, +got error:\n%s", diff)
 							}
 							return nil
@@ -146,7 +147,7 @@ func TestAPICallbacks_Destroy(t *testing.T) {
 					},
 					Scheme: xpfake.SchemeWith(&fake.Terraformed{}),
 				},
-				err: tjerrors.NewDestroyFailed(errBoom.Error()),
+				err: tjerrors.WrapDestroyFailed(errBoom, nil),
 			},
 		},
 		"DestroyOperationSucceeded": {
--- a/pkg/terraform/errors/errors_test.go
+++ b/pkg/terraform/errors/errors_test.go
@@ -1,0 +1,227 @@
+/*
+Copyright 2021 The Crossplane Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package errors
+
+import (
+	"testing"
+
+	"github.com/google/go-cmp/cmp"
+	"github.com/pkg/errors"
+)
+
+var (
+	errorLog = []byte(`{"@level":"info","@message":"Terraform 1.0.3","@module":"terraform.ui","@timestamp":"2021-11-14T23:23:14.009380+03:00","terraform":"1.0.3","type":"version","ui":"0.1.0"}
+{"@level":"error","@message":"Error: Missing required argument","@module":"terraform.ui","@timestamp":"2021-11-14T23:23:14.576254+03:00","diagnostic":{"severity":"error","summary":"Missing required argument","detail":"The argument \"location\" is required, but no definition was found.","range":{"filename":"main.tf.json","start":{"line":24,"column":7,"byte":568},"end":{"line":24,"column":8,"byte":569}},"snippet":{"context":"resource.azurerm_resource_group.example","code":"      }","start_line":24,"highlight_start_offset":6,"highlight_end_offset":7,"values":[]}},"type":"diagnostic"}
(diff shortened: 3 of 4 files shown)


pkg/terraform/errors/errors.go
```diff
@@ -16,40 +16,156 @@ limitations under the License.
...
+
+	parseError, tfError := newTFError(message, cause, logs)
+	if parseError == "" {
+		return tfError
+	}
+	return errors.WithMessage(tfError, parseError)
+}
+
+// WrapApplyFailed returns a new apply failure error with given logs.
+func WrapApplyFailed(cause error, logs []byte) error {
```
muvaf, line 130: I think _wrapping_ in std library means adding information about the upper level in the call stack, however, here we're using logs which is the actual content of the error. In fact, the error returned from `cmd.CombinedOutput()` is usually just `exited with 1` which is not really the cause.

So, I'd suggest treating this as creation of a new error instead of wrapping the command return code with logs. In fact, accepting only `logs` could be even simpler and still give us all the information we need.

```suggestion
func NewApplyFailed(cause error, logs []byte) error {
```

or 

```suggestion
func NewApplyFailed(logs []byte) error {
```

pkg/terraform/errors/errors.go
```diff
@@ -16,40 +16,156 @@ limitations under the License.
...
+		if err := jsoniter.ConfigCompatibleWithStandardLibrary.UnmarshalFromString(l, log); err != nil {
+			return nil, err
+		}
+		tfLogs = append(tfLogs, log)
+	}
+	return tfLogs, nil
+}
+
+// WrapTFError returns a new Terraform CLI failure error with given logs.
+func WrapTFError(message string, cause error, logs []byte) error {
```
muvaf, line 117: Seems like we can have `WrapPlanFailed` and `WrapRefreshFailed` instead of this generic one since command types we run are determined and is a very limited set.

pkg/terraform/errors/errors.go
```diff
@@ -16,40 +16,156 @@ limitations under the License.
...
+	}
+
+	messages := make([]string, 0, len(t.tfLogs))
+	for _, l := range t.tfLogs {
+		// only use error logs
+		if l == nil || l.Level != levelError {
+			continue
+		}
+		m := l.Message
+		if l.Diagnostic.Severity == levelError && l.Diagnostic.Summary != "" {
```
muvaf, line 69: Do you think keeping any information other than what we're printing here has benefits over-weighing the additional logic complexity? I can see us saving a single error string in `tfError` filtered with these conditions in `parseTerraformLogs` and just print that when `Error()` is called, i.e. `applyFailed` having only `message string` field with the message that contains everything we'd like to print.

pkg/terraform/errors/errors.go
```diff
@@ -16,40 +16,156 @@ limitations under the License.
...
 
-// NewApplyFailed returns a new apply failure error with given logs.
-func NewApplyFailed(log string) error {
-	return &applyFailed{log: log}
+func parseTerraformLogs(logs []byte) ([]*TerraformLog, error) {
+	logLines := strings.Split(string(logs), "\n")
+	tfLogs := make([]*TerraformLog, 0, len(logLines))
+	for _, l := range logLines {
+		log := &TerraformLog{}
+		l := strings.TrimSpace(l)
```
muvaf, line 104: Is there any metadata information we can use to filter out the information from HCL? In the past, we were seeing the creds printed in errors and removed them from HCL but if there is a way to not include that at this level, I think we can remove `Env []string` functionality we added back then to simplify things.

muvaf (commented):
Thanks @ulucinar ! My general comment is that there are only a limited set of errors, shapes and even error content that we expect this library to be used for and I believe we should take advantage of this as much as possible by reducing the logic to handle only those instead of making it generic for future possibilities. Both approaches are valid but if we start with the smallest possible set, we won't have to pay the cost of generic-ness until we actually need it.

pkg/terraform/errors/errors.go
ulucinar: The idea here is that we are wrapping the error returned from the `exec` library. Also because I wanted to use these two error types for apply and destroy and the generic tf error type for all operations on a workspace, we were already wrapping errors returned from the `exec` library. For instance, in the previous implementation, we were wrapping as follows in `Refresh`:
```go
	out, err := cmd.CombinedOutput()
	...
	if err != nil {
		return RefreshResult{}, errors.Wrapf(err, "cannot refresh: %s", string(out))
	}
```

Now it becomes:
```go
	out, err := cmd.CombinedOutput()
	...
	if err != nil {
		return RefreshResult{}, tferrors.WrapTFError("cannot refresh", err, out)
	}
```

And I like the idea of wrapping as much as possible as it pertains info about the context.

ulucinar: Another idea here is we would like to generate a TF error only if there is an underlying error. So, converting these to wrappers is more appropriate IMHO. If other use cases arise for these error types, we can then consider adding plain constructors but our current use cases are only wrappers.

pkg/terraform/errors/errors.go
ulucinar: Yes, correct. I did not see any current use case for specialized errors for them but it should be straightforward.

pkg/terraform/errors/errors.go
ulucinar: I had preferred this so that computation of the actual (filtered) error message is deferred until `Error()` is called. But not a big deal in my opinion.

pkg/terraform/errors/errors.go
ulucinar: I would propose to see if we hit that situation again. Previously, we were supplying the provider credentials via the Terraform configuration file (main.hcl) which had caused that issue. Do we know any other cases which could cause a similar issue. If so, I would suggest investigating it in a separate issue maybe.

muvaf: > Previously, we were supplying the provider credentials via the Terraform configuration file (main.hcl) which had caused that issue.

I vaguely recall that the error was printed as one type of JSON and then HCL was printed in another JSON with different classification. What I'm proposing is that if we're able to differentiate programmatically by checking the properties of those JSONs, it'd be great to do it so that we revert to writing credentials to HCL to simplify that part of the codebase. If all was in a single error message that we can't differentiate, then there isn't much we can do.

pkg/terraform/errors/errors.go
muvaf: > And I like the idea of wrapping as much as possible as it pertains info about the context.

The thing is even though it looks like that, it doesn't actually pertains any info about the context in practice; it's only about what error code is returned. I think wrapping would've made more sense if the `err` returned from `exec` actually returned with some information like output of `stderr` for example. However, what it conveys is only the fact that the execution has failed (`exited with code 1`) and we process that information with `err != nil` anyway. So even though it looks usual error wrapping, the error returned is actually misleading and we need to look at the `stderr` and/or `stdout` in all cases. If you think the error code is valuable we can preprend it to the log like `fmt.Sprintf("%s: %s", err.Error(), log)`.

pkg/terraform/errors/errors.go
muvaf: Yeah I think the computations are negligible that the earlier and simpler we can get away, the better.

pkg/terraform/errors/errors.go
turkenh: I believe env solution fixing more than just error messages and we should not revert it: https://github.com/crossplane-contrib/terrajet/issues/99#issuecomment-942033678

ulucinar: Opened #151 to track credential consumption discussion. @muvaf @turkenh, extended the logs with the file name (summary and detail were already there) as we have discussed offline.

muvaf: Thank you @ulucinar !

pkg/terraform/errors/errors.go
```diff
@@ -16,40 +16,169 @@ limitations under the License.
...
+	FileName string `json:"filename"`
+}
+
+func (t *tfError) Error() string {
+	return t.message
 }
 
-func (a *applyFailed) Error() string {
-	return a.log
+func newTFError(message string, logs []byte) (string, *tfError) {
```
muvaf, line 64: nitpitck
```suggestion
func newTFError(message string, logs []byte) (*tfError, error) {
```

pkg/terraform/errors/errors.go
```diff
@@ -16,40 +16,169 @@ limitations under the License.
...
+		return result
+	}
+	return errors.WithMessage(result, parseError)
 }
 
 // IsApplyFailed returns whether error is due to failure of an apply operation.
 func IsApplyFailed(err error) bool {
-	_, ok := err.(*applyFailed)
-	return ok
+	r := &applyFailed{}
```
muvaf, line 122: nit: now it's possible to get away with simple type casting.

muvaf (approved):
Thanks @ulucinar !

pkg/terraform/errors/errors.go
ulucinar: Because `tfError` is shared (embedded) inside actual error types and because we need to consume the returned error instance as a `tfError`, doing this will necessitate type cases at those sites. I'd prefer to keep it as it's as it's not part of the public interface.

pkg/terraform/errors/errors.go
ulucinar: This I also considered. However, in case these errors are themselves wrapped, then the current implementations are more robust. In my opinion, as a principle, we had better always implement similar methods with wrapping in mind.

PR #144, opened 2021-11-15
Rename Group to ShortGroup and allow empty short group

muvaf:

`Group` is actually the full API group like `ec2.aws.crossplane.io` but it seems like the content we're expecting there is not of that form, so I changed its name. Also, I added a logic to allow `""` short group so that some resources can be in provider level group, like `ResourceGroup` in Azure where it should be in `tfazure.crossplane.io`.

I have:

N/A

### DIFF
--- a/pkg/config/provider.go
+++ b/pkg/config/provider.go
@@ -52,13 +52,17 @@ type Provider struct {
 	// for "aws_rds_cluster", we drop "aws_" prefix and its group ("rds") to set
 	// Kind of the resource as "Cluster".
 	TerraformResourcePrefix string
-	// GroupSuffix is the suffix to append to resource groups, e.g.
-	// ".aws.tf.crossplane.io". Defaults to ".<prefix>.tf.crossplane.io".
-	GroupSuffix string
+
+	// RootGroup is the root group that all CRDs groups in the provider are based
+	// on, e.g. "aws.tf.crossplane.io".
+	// Defaults to "<TerraformResourcePrefix>.tf.crossplane.io".
+	RootGroup string
+
 	// ShortName is the short name of the provider. Typically, added as a CRD
 	// category, e.g. "tfaws". Default to "tf<prefix>". For more details on CRD
 	// categories, see: https://kubernetes.io/docs/tasks/extend-kubernetes/custom-resources/custom-resource-definitions/#categories
 	ShortName string
+
 	// ModulePath is the go module path for the tf based provider repo, e.g.
 	// "github.com/crossplane-contrib/provider-tf-aws"
 	ModulePath string
@@ -77,6 +81,7 @@ type Provider struct {
 	// can add "aws_shield_protection_group$". To skip whole aws waf group, one
 	// can add "aws_waf.*" to the list.
 	SkipList []string
+
 	// IncludeList is a list of regex for the Terraform resources to be
 	// skipped. For example, to include "aws_shield_protection_group" into
 	// the generated resources, one can add "aws_shield_protection_group$".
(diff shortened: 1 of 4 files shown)


pkg/config/provider.go
```diff
@@ -52,9 +52,9 @@ type Provider struct {
 	// for "aws_rds_cluster", we drop "aws_" prefix and its group ("rds") to set
 	// Kind of the resource as "Cluster".
 	TerraformResourcePrefix string
-	// GroupSuffix is the suffix to append to resource groups, e.g.
-	// ".aws.tf.crossplane.io". Defaults to ".<prefix>.tf.crossplane.io".
-	GroupSuffix string
+	// RootGroup is the suffix to append to resource groups, e.g.
+	// "aws.tf.crossplane.io". Defaults to "<prefix>.tf.crossplane.io".
```
ulucinar, line 56: Here, I guess we are referring to the `prefix` argument passed to `config.NewProvider` with `<prefix>`.  We may make this clear.

pkg/config/resource.go
```diff
@@ -160,8 +160,12 @@ type Resource struct {
...
 	IDFieldName string
 
-	// Group is the group of CRD.
-	Group string
+	// ShortGroup is the short name of the API group of this CRD. The full
+	// CRD API group is calculated by adding the group suffix of the provider.
+	// For example, ShortGroup could be `ec2` where group suffix of the
+	// provider is `aws.crossplane.io` and in that case, the full group would
+	// be `ec2.aws.crossplane.io`
+	ShortGroup string
```
ulucinar, line 168: `ShortGroup` is fine but we can also consider something like `GroupPrefix` as its name. From a "Short" version, I expect something that _substitutes_ a longer version of it, like in the context of command-line options, "-d" being a substitute for "--debug".

pkg/pipeline/run.go
```diff
@@ -37,13 +37,13 @@ func Run(pc *config.Provider, rootDir string) { // nolint:gocyclo
 	// Group resources based on their Group and API Versions.
 	resourcesGroups := map[string]map[string]map[string]*config.Resource{}
```
ulucinar, line 38: A more explicit comment would help here, maybe using a syntax like:
`<API group prefix/short group>` -> <API version> -> <resource name>` and maybe with a specific example for a resource.

pkg/pipeline/run.go
```diff
@@ -57,12 +57,16 @@ func Run(pc *config.Provider, rootDir string) { // nolint:gocyclo
 		controllerPkgList = append(controllerPkgList, filepath.Join(pc.ModulePath, p))
 	}
 	count := 0
-	for group, versions := range resourcesGroups {
+	for shortGroup, versions := range resourcesGroups {
+		group := pc.RootGroup
+		if shortGroup != "" {
```
ulucinar, line 62: We can consider trimming before comparison and concatenation.

ulucinar (approved):
Thanks @muvaf for the PR. I have consumed these changes in https://github.com/crossplane-contrib/provider-tf-azure/pull/83. Below are some minor comments for your consideration.

pkg/config/resource.go
muvaf: Yes, in fact, it substitutes the longer version, i.e. `ec2` for `ec2.aws.crossplane.io`. But I don't love `ShortGroup` either. `Group` only is misleading because the actual group is `ec2.aws.crossplane.io`.

The main motivation was that I didn't want to treat those two fields (`GroupPrefix` and `GroupSuffix`) as pure strings with those names because the concatenation we're doing now is not simply prepending anymore; we're opening  a new group under the root group if short group is given. Open to suggestions for the name.

pkg/pipeline/run.go
muvaf: As synced, we can treat this as developer error if there is space there.

PR #145, opened 2021-11-16
Use explicit root directory so that generator can be run from anywhere

muvaf:

This change lets the developer of the provider choose how they take the root dir by not enforcing an implicit contract by Terrajet like having to run it from the top level directory.

The main reason I wanted to do this is that I'd like to run `go run ../cmd/generator/main.go` from `apis/generate.go` and that won't work because the license header path is calculated as relative path.

I have:

Manually with provider-tf-aws.


turkenh (approved):
lgtm!

PR #149, opened 2021-11-17
Use pointer for secret ref if optional

### LINKED ISSUE #118: Optional Secret References Are Always Resolved
### What happened?

While working on the `azurerm_sql_server` resource, we have the situation where a secret reference is optional but we try to resolve a zero-valued reference and failing with a secret-not-found error. Here are some details:
In Terraform azurerm provider, SQL server's `threat_detection_policy[*]. storage_account_access_key` field is only required if the policy is enabled, and thus is marked as optional in its [Terraform schema](https://github.com/hashicorp/terraform-provider-azurerm/blob/c93c339c8ca1a35efe4015bc174561feb814e6ce/internal/services/sql/sql_server_resource.go#L179). And for a *valid* disabled policy with no reference to a storage account access key, we still try to resolve a zero-valued secret reference and fail to do so. The reported error message in status.conditions is:
```
'connect failed: cannot get a terraform workspace for resource: cannot
      create a new file producer: cannot get sensitive parameters: cannot get secret
      value for &{{ } }: cannot get secret data: Secret "" not found'
```

### How can we reproduce it?

To be added.

turkenh:

Fixes #118

I have:

Tested with gcp compute instance which also has the same problem described in #118 

### DIFF
--- a/pkg/types/builder.go
+++ b/pkg/types/builder.go
@@ -156,11 +156,14 @@ func (g *Builder) buildResource(res *schema.Resource, cfg *config.Resource, tfPa
 			// If it is an observation field, it will be dropped.
 			// Data will be loaded from the referenced secret key.
 			fieldNameCamel += sfx
-			// todo(hasan): do we need the pointer type if optional?
-			fieldType = typeSecretKeySelector
 
-			jsonTag = fmt.Sprintf("%s,omitempty", NewNameFromCamel(fieldNameCamel).LowerCamelComputed)
 			tfTag = "-"
+			fieldType = typeSecretKeySelector
+			jsonTag = NewNameFromCamel(fieldNameCamel).LowerCamelComputed
+			if sch.Optional {
+				fieldType = types.NewPointer(typeSecretKeySelector)
+				jsonTag += ",omitempty"
+			}
 		}
 		field := types.NewField(token.NoPos, g.Package, fieldNameCamel, fieldType, false)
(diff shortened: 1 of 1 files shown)


pkg/types/builder.go
```diff
@@ -156,11 +156,14 @@ func (g *Builder) buildResource(res *schema.Resource, cfg *config.Resource, tfPa
...
 			// Data will be loaded from the referenced secret key.
 			fieldNameCamel += sfx
-			// todo(hasan): do we need the pointer type if optional?
-			fieldType = typeSecretKeySelector
 
-			jsonTag = fmt.Sprintf("%s,omitempty", NewNameFromCamel(fieldNameCamel).LowerCamelComputed)
 			tfTag = "-"
+			fieldType = typeSecretKeySelector
+			jsonTag = fmt.Sprintf("%s", NewNameFromCamel(fieldNameCamel).LowerCamelComputed)
+			if sch.Optional {
```
muvaf, line 163: Don't we need to skip [here](https://github.com/crossplane-contrib/terrajet/blob/6d7fb7e/pkg/resource/sensitive.go#L167) if the value of that pointer is `nil`?

turkenh: No, because expandWildcards method [here](https://github.com/crossplane-contrib/terrajet/blob/6d7fb7e/pkg/resource/sensitive.go#L162), does only return paths that are available in the spec.

muvaf (approved):
Thanks @turkenh !

PR #159, opened 2021-11-24
Allow prevent_destroy meta-arg to be overridden per resource

### LINKED ISSUE #158: Support per-Resource `lifecycle.prevent_destroy` meta-arg Configuration

For certain Terraform resources like Azure's PostgreSQL server configuration, we should not be including the `lifecycle.prevent_destroy` meta-arg in order to properly implement updates. Refer to https://github.com/crossplane-contrib/provider-jet-azure/issues/97 for one such example.

We could extend our resource configuration framework with an `AllowDestroy` parameter.

ulucinar:

Fixes #158

For certain Terraform resources like Azure's PostgreSQL server configuration, we should not be including the `lifecycle.prevent_destroy` meta-arg for updates. This PR proposes a backwards-compatible change that allows us to override this behavior. The default behavior stays the same, unless explicitly configured for a resource, we always include (apart from deletions) the `prevent_destroy: true` meta-argument in the generated configuration.

I have:

Tested using the example manifest of PostgreSQLConfiguration resource of `provider-jet-azure` on top of https://github.com/crossplane-contrib/provider-jet-azure/pull/98.

### DIFF
--- a/pkg/config/resource.go
+++ b/pkg/config/resource.go
@@ -212,6 +212,12 @@ type Resource struct {
 	// databases.
 	UseAsync bool
 
+	// AllowDestroy controls whether the `prevent_destroy: true` stanza
+	// is included in the Terraform configuration for the resource.
+	// For certain resources like Azure's PostgreSQLConfiguration,
+	// we would like to allow replacement.
+	AllowDestroy bool
+
 	// ExternalName allows you to specify a custom ExternalName.
 	ExternalName ExternalName
 
--- a/pkg/controller/external.go
+++ b/pkg/controller/external.go
@@ -181,7 +181,7 @@ func (e *external) Observe(ctx context.Context, mg xpresource.Managed) (managed.
 
 	return managed.ExternalObservation{
 		ResourceExists:          true,
-		ResourceUpToDate:        plan.UpToDate,
+		ResourceUpToDate:        plan.UpToDate && plan.Exists,
 		ResourceLateInitialized: lateInitedAnn || lateInitedParams,
 		ConnectionDetails:       conn,
 	}, nil
--- a/pkg/terraform/files.go
+++ b/pkg/terraform/files.go
@@ -148,7 +148,7 @@ func (fp *FileProducer) WriteMainTF() error {
 	// If the resource is in a deletion process, we need to remove the deletion
 	// protection.
 	fp.parameters["lifecycle"] = map[string]bool{
-		"prevent_destroy": !meta.WasDeleted(fp.Resource),
+		"prevent_destroy": !meta.WasDeleted(fp.Resource) && !fp.Config.AllowDestroy,
 	}
 	// Note(turkenh): To use third party providers, we need to configure
 	// provider name in required_providers.


muvaf (changes requested):
Please see https://github.com/crossplane-contrib/provider-jet-azure/issues/97#issuecomment-978124981 .

pkg/controller/external.go
```diff
@@ -181,7 +181,7 @@ func (e *external) Observe(ctx context.Context, mg xpresource.Managed) (managed.
 
 	return managed.ExternalObservation{
 		ResourceExists:          true,
-		ResourceUpToDate:        plan.UpToDate,
+		ResourceUpToDate:        plan.UpToDate && plan.Exists,
```
muvaf, line 184: Is this really necessary?

pkg/config/resource.go
```diff
@@ -212,6 +212,12 @@ type Resource struct {
 	// databases.
 	UseAsync bool
 
+	// AllowDestroy controls whether the `prevent_destroy: true` stanza
+	// is included in the Terraform configuration for the resource.
+	// For certain resources like Azure's PostgreSQLConfiguration,
+	// we would like to allow replacement.
+	AllowDestroy bool
```
muvaf, line 219:
```suggestion
	// AllowRecreation controls whether the changes that require resource
	// re-creation, i.e. immutable fields, should be allowed, which is against
	// Crossplane Resource Model.
	//
	// This should be enabled only in cases where there is no point in lifecycle
	// of the resource where the API returns NotFound. For example, the
	// configurations that Azure PostgreSQLConfiguration resource changes
	// have defaults, hence they always exist and when you create a new
	// PostgreSQLConfiguration, it's always an importing process of the default
	// so the operation would have to be a re-creation since all parameters are
	// immutable. Otherwise, controller imports instead of creation since the
	// resource exists and ends up in dead-lock state since re-creation is no
	// allowed.
	//
	// For other cases, enabling this is strictly against Crossplane Resource Model
	// where the deletion of the external resource is issued if and only if the custom
	// resource in Kubernetes API is deleted. Use this parameter with high caution.
	AllowRecreation bool
```

muvaf (commented):
@ulucinar I [understand](https://github.com/crossplane-contrib/provider-jet-azure/issues/97#issuecomment-978201436) the necessity of this change now. I added a long comment section in the review to make it more clear but while I was writing it, I had some second thoughts because while the use case we have doesn't, the concept itself allows something that is very much against Crossplane conventions (deleting without user consent given we don't have interactive confirmation like TF does before taking action), which could cause developers to use it with not as much caution.

What would you think about excluding these resources (`PostgreSQLConfiguration`, `MySQLConfiguration` and possibly `MSSQLConfiguration`) from Jet Azure and let people use the implementation in provider-azure? I'd like us to see more resources with that exception so that it's worth handling at the top level and take the risk.

muvaf:
@ulucinar Another alternative could be to copy the implementation from the native provider. All the structure and flows of Jet providers are just like native implementation, so it should work OK if you'd prefer to have those resources in Jet Azure.

turkenh:
Should we close this one (for now) given the [original issue](https://github.com/crossplane-contrib/provider-jet-azure/issues/97) resolved without needing this?

PR #160, opened 2021-11-26
Make Terraform resource ID available in status.atProvider

### LINKED ISSUE #154: Make Terraform resource ID available in managed resource

For certain providers, like `provider-jet-azure`, we need to be able to resolve cross-resource references to the provider's assigned ID for that resource. For instance in Azure context, this corresponds to a resolution to:
`/subscriptions/00000000-0000-0000-0000-000000000000/resourcegroups/group1/providers/Microsoft.ContainerService/managedClusters/cluster1`

Not all pieces of this information is available in the Terraformed MR. When crossplane-runtime tries to resolve a reference to a `KubernetesCluster` object, currently we cannot perform this solely from the MR.

We could have Terrajet inject the ID meta-attribute, which is available in the Terraform state, into the MR via an annotation or via the means of a `Terraformed.Status` struct, which embeds crossplane-runtime's `Status` and extends it with an ID.

ulucinar:
Signed-off-by: Alper Rifat Ulucinar <ulucinar@users.noreply.github.com>

Fixes #154 

This PR proposes a change to make the persistent ID available in the Terraform state part of `status.atProvider` of the generated managed resource.

I have:

### DIFF
--- a/pkg/pipeline/crd.go
+++ b/pkg/pipeline/crd.go
@@ -23,6 +23,7 @@ import (
 	"path/filepath"
 	"strings"
 
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
 	twtypes "github.com/muvaf/typewriter/pkg/types"
 	"github.com/muvaf/typewriter/pkg/wrapper"
 	"github.com/pkg/errors"
@@ -66,6 +67,12 @@ func (cg *CRDGenerator) Generate(cfg *config.Resource) (string, error) {
 	for _, omit := range cfg.ExternalName.OmittedFields {
 		delete(cfg.TerraformResource.Schema, omit)
 	}
+
+	cfg.TerraformResource.Schema["id"] = &schema.Schema{
+		Type:     schema.TypeString,
+		Computed: true,
+	}
+
 	gen, err := tjtypes.NewBuilder(cg.pkg).Build(cfg)
 	if err != nil {
 		return "", errors.Wrapf(err, "cannot build types for %s", cfg.Kind)
--- a/pkg/pipeline/templates/terraformed.go.tmpl
+++ b/pkg/pipeline/templates/terraformed.go.tmpl
@@ -55,6 +55,11 @@ func (tr *{{ .CRD.Kind }}) SetObservation(obs map[string]interface{}) error {
 	return json.TFParser.Unmarshal(p, &tr.Status.AtProvider)
 }
 
+// GetObservation of this {{ .CRD.Kind }}
+func (tr *{{ .CRD.Kind }}) GetID() string {
+    return tr.Status.AtProvider.ID
+}
+
 // GetParameters of this {{ .CRD.Kind }}
 func (tr *{{ .CRD.Kind }}) GetParameters() (map[string]interface{}, error) {
 	p, err := json.TFParser.Marshal(tr.Spec.ForProvider)
--- a/pkg/pipeline/templates/terraformed.go.tmpl
+++ b/pkg/pipeline/templates/terraformed.go.tmpl
@@ -38,6 +38,16 @@ func (tr *{{ .CRD.Kind }}) GetObservation() (map[string]interface{}, error) {
 
 // SetObservation for this {{ .CRD.Kind }}
 func (tr *{{ .CRD.Kind }}) SetObservation(obs map[string]interface{}) error {
+    id, ok := obs["id"]
+    if !ok {
+        return errors.New("attribute id not found in observed state")
+    }
+    idStr, ok := id.(string)
+    if !ok {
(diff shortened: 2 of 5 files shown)


pkg/pipeline/templates/terraformed.go.tmpl
```diff
@@ -38,6 +38,16 @@ func (tr *{{ .CRD.Kind }}) GetObservation() (map[string]interface{}, error) {
...
 func (tr *{{ .CRD.Kind }}) SetObservation(obs map[string]interface{}) error {
+    id, ok := obs["id"]
+    if !ok {
+        return errors.New("attribute id not found in observed state")
+    }
+    idStr, ok := id.(string)
+    if !ok {
+        return errors.New("attribute id in observed state is not of type string")
+    }
+    tr.Status.AtProvider.ID = idStr
```
muvaf, line 49: What do you think about having this under status directly, i.e. `tr.Status.ID`? I'm worried that we'll see many conflicts in the observation since `id` is a common field name.

pkg/pipeline/templates/terraformed.go.tmpl
```diff
@@ -38,6 +38,16 @@ func (tr *{{ .CRD.Kind }}) GetObservation() (map[string]interface{}, error) {
 
 // SetObservation for this {{ .CRD.Kind }}
 func (tr *{{ .CRD.Kind }}) SetObservation(obs map[string]interface{}) error {
```
muvaf, line 40: `GetID() string` could be a pretty useful addition to `resource.Terraformed` interface that will allow us use it in references without marshaling.

pkg/pipeline/templates/terraformed.go.tmpl
muvaf: Maybe another name, like `TerrajetID` or `JetID`?

pkg/pipeline/templates/terraformed.go.tmpl
muvaf: Actually it'd be a nice-to-have since we can cast to the exact type without marshaling anyway, just like in native providers. Perfectly fine to not add it if you don't want to make a change to that interface.

pkg/pipeline/templates/terraformed.go.tmpl
ulucinar: As we have discussed in the sync, my observation is that the `id` is in the set of returned attributes for the Terraform state, and hence I don't expect a Terraform resource schema to define a conflicting attribute with it. As an example:
```hcl
 "resources": [
    {
      "mode": "managed",
      "type": "azurerm_resource_group",
      "name": "example",
      "provider": "provider[\"registry.terraform.io/hashicorp/azurerm\"]",
      "instances": [
        {
          "schema_version": 0,
          "attributes": {
            "id": "/subscriptions/.../resourceGroups/example-rg-alper",
            "location": "eastus",
            "name": "example-rg-alper",
            "tags": null,
            "timeouts": null
          },
...
```

And because this is the persistent ID of the resource as observed on the Cloud provider side, I had decided to put this into `status.atProvider`. But I had also considered making it directly available at `status`. I'm also fine with it. Just semantically I think it makes sense at `status.atProvider`.

pkg/pipeline/templates/terraformed.go.tmpl
ulucinar: I think it's fine to add considering this interface is not in broad use in the wild. If agreed, I think we should be fine to introduce a backward-incompatible change at this maturity level.

pkg/pipeline/crd.go
```diff
@@ -66,6 +67,12 @@ func (cg *CRDGenerator) Generate(cfg *config.Resource) (string, error) {
 	for _, omit := range cfg.ExternalName.OmittedFields {
 		delete(cfg.TerraformResource.Schema, omit)
 	}
+
+	cfg.TerraformResource.Schema["id"] = &schema.Schema{
```
muvaf, line 71: As we discussed in sync, `id` field is [integral part](https://learn.hashicorp.com/tutorials/terraform/provider-create#implement-create) (search for `SetId`) of the state because Terraform assumes that resource does not exist if `id` is not set. So, it should be safe to always add it to the schema.

muvaf (approved):
Thanks @ulucinar !

pkg/pipeline/crd.go
ulucinar: I agree. One point to keep in mind is that we have some further assumptions like in the schema, it's computed.

muvaf: At that level, I think we can add a nil guard, like `if cr.Status.AtProvider.ID == nil { return "" }` in `GetID() string`. My point was that we are sure that field is reserved for `id`, so we don't have to check whether it comes with the schema.

PR #162, opened 2021-11-29
Add fake utility for cli calls

### LINKED ISSUE #81: Mock utility for command call tests

We use `exec` package to call Terraform CLI and there isn't a utility like `afero` to replace it with test structs. So, it's really cumbersome to test functions that make CLI calls.

We can implement a utility similar to [`afero`](https://github.com/spf13/afero) but for CLI calls.

sergenyalcin:

Fixes #81 

This PR adds a utility for the command calls. There are two interfaces to manage the command and command outputs. 

Mock tests can be written by using this utility. `StdErr` and `StdOut` values can be manipulated via `FakeExec` struct that implements the Fake call of the `CombinedOutput` function.

In [real CombinedOutput] function, the StdErr and StdOut values are assigned to the same bytes.Buffer instance. Therefore, I also tried to implement the same logic in this [PR].

I have:

This PR was tested by adding a toy test to observe the fake functions are working properly.

[real CombinedOutput]: https://github.com/golang/go/blob/1ea4d3b91164fb08b7022958b6cd8e290f12e017/src/os/exec/exec.go#L564
[PR]: https://github.com/sergenyalcin/terrajet/blob/11db5953e1b20eb63581b5e0dc584dc35d851a3c/pkg/terraform/exec.go#L48

### DIFF
--- a/pkg/controller/external.go
+++ b/pkg/controller/external.go
@@ -26,6 +26,7 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/client"
 
 	"github.com/crossplane-contrib/terrajet/pkg/config"
+	"github.com/crossplane-contrib/terrajet/pkg/exec"
 	"github.com/crossplane-contrib/terrajet/pkg/resource"
 	"github.com/crossplane-contrib/terrajet/pkg/resource/json"
 	"github.com/crossplane-contrib/terrajet/pkg/terraform"
@@ -98,7 +99,7 @@ func (c *Connector) Connect(ctx context.Context, mg xpresource.Managed) (managed
 		return nil, errors.Wrap(err, errGetWorkspace)
 	}
 
-	tf.Executor = &terraform.RealExec{}
+	tf.Executor = &exec.OsExec{}
 
 	return &external{
 		workspace: tf,
--- a/pkg/controller/external.go
+++ b/pkg/controller/external.go
@@ -98,6 +98,8 @@ func (c *Connector) Connect(ctx context.Context, mg xpresource.Managed) (managed
 		return nil, errors.Wrap(err, errGetWorkspace)
 	}
 
+	tf.Executor = &terraform.RealExec{}
+
 	return &external{
 		workspace: tf,
 		config:    c.config,
--- a/pkg/terraform/exec.go
+++ b/pkg/terraform/exec.go
@@ -1,0 +1,97 @@
+/*
+Copyright 2021 The Crossplane Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package terraform
+
+import (
+	"bytes"
+	"context"
+	"os"
+	"os/exec"
+)
+
+// Exec interface provides a Command interface. By this way, real or fake implementation of functions is determined.
+type Exec interface {
+	CommandContext(ctx context.Context, name string, dir string, env []string, arg ...string) Command
+}
+
+// RealExec struct implements the CommandContext function that calls the function from the exec package.
+type RealExec struct{}
+
+// CommandContext calls the function from exec package, then configure the *Cmd object with `dir` and `env`.
+// It returns a *Cmd object thus, the real implementation of CombinedOutput will be called.
+func (r *RealExec) CommandContext(ctx context.Context, name string, dir string, env []string, arg ...string) Command {
+	cmd := exec.CommandContext(ctx, name, arg...)
+
+	configureCmd(cmd, dir, env)
+
+	return cmd
+}
+
+// FakeExec struct implements the CommandContext function that calls fake one.
+type FakeExec struct {
+	stdErr []byte
+	stdOut []byte
+	err    error
+}
+
+// CommandContext function manipulates the `StdErr` and `stdOut` .
+// It returns a FakeCommand object thus, the fake implementation of CombinedOutput will be called.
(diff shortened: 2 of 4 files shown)


pkg/terraform/exec.go
```diff
@@ -0,0 +1,97 @@
...
+	"os/exec"
+)
+
+// Exec interface provides a Command interface. By this way, real or fake implementation of functions is determined.
+type Exec interface {
+	CommandContext(ctx context.Context, name string, dir string, env []string, arg ...string) Command
+}
+
+// RealExec struct implements the CommandContext function that calls the function from the exec package.
+type RealExec struct{}
```
muvaf, line 32:
```suggestion
type StdExec struct{}
```
or
```suggestion
type OSExec struct{}
```
nitpick

pkg/terraform/exec.go
```diff
@@ -0,0 +1,97 @@
...
+	}
+}
+
+// Command interface
+type Command interface {
+	CombinedOutput() ([]byte, error)
+}
+
+// FakeCommand struct implements the CombinedOutput function that calls fake one.
+type FakeCommand struct {
```
muvaf, line 80: An initializer with [functional options](https://github.com/crossplane/crossplane/blob/master/CONTRIBUTING.md#dont-wrap-function-signatures) could be pretty useful here, i.e. `NewFakeCommand(opts ...FakeCommandOption) *FakeCommand`. Then users can call `NewFakeCommand(exec.WithStdOut(string), exec.WithError(err))` etc.

pkg/controller/external.go
```diff
@@ -98,6 +98,8 @@ func (c *Connector) Connect(ctx context.Context, mg xpresource.Managed) (managed
 		return nil, errors.Wrap(err, errGetWorkspace)
 	}
 
+	tf.Executor = &terraform.RealExec{}
```
muvaf, line 101: This seems very `terraform` package specific. What do you think about doing this in `WorkspaceStore`? i.e. by `WorkspaceStore` having a field accepting `Exec` type.

muvaf (commented):
Going great @sergenyalcin !

pkg/terraform/exec.go
muvaf: or `Std`/`OS` if we put them under `exec` package.

pkg/terraform/exec.go
```diff
@@ -0,0 +1,97 @@
...
+	cmd := exec.CommandContext(ctx, name, arg...)
+
+	configureCmd(cmd, dir, env)
+
+	return cmd
+}
+
+// FakeExec struct implements the CommandContext function that calls fake one.
+type FakeExec struct {
+	stdErr []byte
```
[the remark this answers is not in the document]
sergenyalcin: Do you mean that removing the stdErr and stdOut fields. I think you’re talking about manipulating these values by replacing them with functions. Like that:

```
type FakeExec struct {
	func... // instead of stdErr
	func... // instead of stdOut
	err    error
}
```

pkg/controller/external.go
sergenyalcin: In functions such as Apply, Destroy, as far as I see, we use the Workspace struct. Therefore, it seems that we will have to pass the Exec interface to Workspace after declaration of WorkspaceStore. Could you clarify a bit so I can understand correctly and address your comment properly? Thanks!

sergenyalcin:
Hi @muvaf thanks for your review. I addressed many of your comments. However, I left a few questions to better understand some of them.

pkg/terraform/exec.go
```diff
@@ -0,0 +1,97 @@
...
+	err error
+}
+
+// CombinedOutput returns the output and error.
+func (c *FakeCommand) CombinedOutput() ([]byte, error) {
+	return c.out, c.err
+}
+
+func configureCmd(cmd *exec.Cmd, dir string, env []string) {
+	if cmd.Env == nil {
```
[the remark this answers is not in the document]
sergenyalcin: I expanded the Exec interface by adding the SetEnv and SetDir funcs. Also for fake and os structs of Exec, added dir and env fields.

pkg/terraform/exec.go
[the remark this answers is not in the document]
muvaf: This exec one can actually be used to initialize more than a single `Command` and users might want to return a different one depending on the arguments. For example, if the command is `terraform`, I can return a `Command` that returns a certain `stdOut` but if it's `cat`, I may want to return something else because these outputs are command-specific rather than exec library-specific. Giving users the ability to provide these differentiations would require a function context.

tldr; we wouldn't have any of those fields, but only a function to override what `CommandContext` does.

pkg/controller/external.go
muvaf: My point was that which `Exec` type we'd like to use is something we decide when we start the controller, i.e. whether it's a test or actual prod usage, very similar to what `Fs` backend we choose for Afero layer.

Here we make this decision in the controller of a specific CR but we don't need to be that specific; we can just accept it as parameter in [`NewWorkspaceStore`](https://github.com/crossplane-contrib/terrajet/blob/8e96bd1/pkg/terraform/store.go#L75).

pkg/terraform/exec.go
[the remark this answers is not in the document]
muvaf: These are command-specific, so they need to be configurable on `Command` level rather than `Exec`. One way to do it is to hold a map in the exec structs that can identify commands created using those execs and go and set the env/dir to each whenever needed but that's probably not worth the complexity. I think we'll need to wrap `*exec.Cmd` with a `Cmd` struct here.

I had [originally thought](https://github.com/crossplane-contrib/terrajet/issues/81#issuecomment-978142154) we may be able to get away with making our interfaces satisfiable by `*exec.Cmd` but it doesn't seem like we can do this unless we bring in that complexity.

pkg/terraform/exec.go
muvaf: Seems like they are in `exec` package now but names still contain exec. Repeating the package name in type/function names is [not considered](https://github.com/golang/go/wiki/CodeReviewComments#package-names) very idiomatic, i.e. instead of `exec.OSExec`, we can use `exec.OS` an it'd still provide enough information.

muvaf:
@sergenyalcin has decided to use an implementation from k8s libraries, which turned out to be very similar to his design, so we don't need this anymore.

PR #170, opened 2021-12-06
Handle sensitive optionals coming as null

### LINKED ISSUE #161: Optional sensitive fields are not supported
### What happened?

First reported in https://github.com/crossplane-contrib/provider-tf-equinix-metal/issues/8, fields that are optional and sensitive in the Terraform provider will not prevent resource creation but will cause errors when the resource is being refreshed - thus preventing deletion.

In the Equinix Metal provider's `metal_device` resource, optional sensitive fields `userdata` and `custom_data` are set on device create are not refreshed on subsequent reads. These fields are not forcenew.

https://github.com/equinix/terraform-provider-metal/blob/main/metal/resource_metal_device.go#L268-L282

In Terraform, these fields may appear to be `null` in the `terraform.state` `resources[*].instances[*].attributes` map.

When encountering these values, Terrajet errors out (with a warning): 
```
Warning  CannotObserveExternalResource  30m                  
managed/device.equinixmetal.jet.crossplane.io/v1alpha1, kind=device  
cannot get connection details: cannot get connection details: 
cannot not get a string for fieldpath "user_data": user_data: not a string
```
https://github.com/crossplane-contrib/terrajet/blob/6abd89a444231a61237788434f38d6e92dbae150/pkg/resource/sensitive.go#L119

### How can we reproduce it?

Attempt to create and then delete the following resource: 
https://github.com/crossplane-contrib/provider-tf-equinix-metal/blob/main/examples/device/device.yaml

Or, attempt to delete any resource with an optional sensitive attribute value that has no assigned value.

turkenh:

Fixes #161

I have:

With `device` resource in provider-tf-equinix-metal and also with Unit tests.

### DIFF
--- a/pkg/resource/sensitive.go
+++ b/pkg/resource/sensitive.go
@@ -34,6 +34,7 @@ import (
 
 const (
 	errCannotExpandWildcards          = "cannot expand wildcards"
+	errFmtCannotGetValueForFieldPath  = "cannot not get a value for fieldpath %q"
 	errFmtCannotGetStringForFieldPath = "cannot not get a string for fieldpath %q"
 )
 
@@ -114,9 +115,18 @@ func GetSensitiveAttributes(from map[string]interface{}, mapping map[string]stri
 		}
 
 		for _, fp := range fieldPaths {
-			v, err := paved.GetString(fp)
+			v, err := paved.GetValue(fp)
 			if err != nil {
-				return nil, errors.Wrapf(err, errFmtCannotGetStringForFieldPath, fp)
+				return nil, errors.Wrapf(err, errFmtCannotGetValueForFieldPath, fp)
+			}
+			// Gracefully skip if v is nil which implies that this field is
+			// optional and not provided.
+			if v == nil {
+				continue
+			}
+			s, ok := v.(string)
+			if !ok {
+				return nil, errors.Errorf(errFmtCannotGetStringForFieldPath, fp)
 			}
 			// Note(turkenh): k8s secrets uses a strict regex to validate secret
 			// keys which does not allow having brackets inside. So, we need to
@@ -130,7 +140,7 @@ func GetSensitiveAttributes(from map[string]interface{}, mapping map[string]stri
 			if vals == nil {
 				vals = map[string][]byte{}
 			}
-			vals[fmt.Sprintf("%s%s", prefixAttribute, k)] = []byte(v)
+			vals[fmt.Sprintf("%s%s", prefixAttribute, k)] = []byte(s)
 		}
 	}
 	return vals, nil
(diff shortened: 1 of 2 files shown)


pkg/resource/sensitive.go
```diff
@@ -114,9 +115,18 @@ func GetSensitiveAttributes(from map[string]interface{}, mapping map[string]stri
 		}
 
 		for _, fp := range fieldPaths {
-			v, err := paved.GetString(fp)
+			v, err := paved.GetValue(fp)
 			if err != nil {
```
muvaf, line 119: We probably need to ignore `NotFound` error here https://github.com/crossplane/crossplane-runtime/blob/47bff13/pkg/fieldpath/paved.go#L39

pkg/resource/sensitive.go
```diff
@@ -160,6 +170,17 @@ func GetSensitiveParameters(ctx context.Context, client SecretClient, from runti
 			return errors.Wrapf(err, "cannot expand wildcard for xp resource")
 		}
 		for _, expandedJSONPath := range jsonPathSet {
+			v, err := pavedJSON.GetValue(expandedJSONPath)
+			if err != nil {
```
muvaf, line 174: Here too https://github.com/crossplane/crossplane-runtime/blob/47bff13/pkg/fieldpath/paved.go#L39

pkg/resource/sensitive_test.go
```diff
@@ -227,6 +231,15 @@ func TestGetSensitiveAttributes(t *testing.T) {
 				},
 			},
 		},
+		"Optional": {
```
muvaf, line 234: Would you mind adding a case where the field does not exist at all, i.e. not even `null`?

pkg/resource/sensitive_test.go
```diff
@@ -417,13 +426,13 @@ func TestGetSensitiveParameters(t *testing.T) {
...
 						Key: "pass",
 					})).Return([]byte("foo"), nil)
-					client.EXPECT().GetSecretValue(gomock.Any(), gomock.Eq(xpv1.SecretKeySelector{
-						SecretReference: xpv1.SecretReference{
-							Name:      "admin-key",
-							Namespace: "crossplane-system",
-						},
-						Key: "key",
-					})).Return([]byte("bar"), nil)
+					/*					client.EXPECT().GetSecretValue(gomock.Any(), gomock.Eq(xpv1.SecretKeySelector{
```
muvaf, line 429: Removing these would be cleaner.

muvaf (commented):
Thanks @turkenh !

pkg/resource/sensitive.go
muvaf: In fact, you may continue to use `GetString` if we ignore the error and check against `""` in the next step, hence remove type casting.

turkenh: @muvaf we have two different cases here:

1. No value at that path (i.e. path does not exist)
2. Path exists but set as `null`/`nil`

First case already covered with https://github.com/crossplane-contrib/terrajet/pull/149 and `NotFound` error is consumed by [expandWildcard](https://github.com/crossplane/crossplane-runtime/blob/d43d510ec5786c455983dede48878b5231de50a8/pkg/fieldpath/paved.go#L191).

This PR is fixing second case where path exists but we are getting a null value.

pkg/resource/sensitive.go
turkenh: See response there.

pkg/resource/sensitive_test.go
turkenh: We already have a test case for that: https://github.com/crossplane-contrib/terrajet/blob/e1a8ff4a3f6c28c31101a95533c54150c1917631/pkg/resource/sensitive_test.go#L230

pkg/resource/sensitive.go
muvaf: Got it, I didn't realize we have `null`s in tfstate. Thanks!

muvaf (approved):
Thanks @turkenh !

PR #172, opened 2021-12-07
Unit tests for terraform.Workspace

### LINKED ISSUE #157: Unit tests for terraform.Workspace and its CLI calls

Implement unit tests for CLI calls

Depends on https://github.com/crossplane-contrib/terrajet/issues/81

sergenyalcin:
Signed-off-by: Sergen Yalçın <yalcinsergen97@gmail.com>

Fixes #157 

Added unit tests for Apply, Destroy, Refresh, Plan and Async functions.

I have:

### DIFF
--- a/pkg/terraform/store.go
+++ b/pkg/terraform/store.go
@@ -19,7 +19,6 @@ package terraform
 import (
 	"context"
 	"os"
-	"os/exec"
 	"path/filepath"
 	"sync"
 
@@ -28,6 +27,7 @@ import (
 	"github.com/pkg/errors"
 	"github.com/spf13/afero"
 	"k8s.io/apimachinery/pkg/types"
+	"k8s.io/utils/exec"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 
 	"github.com/crossplane-contrib/terrajet/pkg/config"
@@ -69,10 +69,11 @@ func WithFs(fs afero.Fs) WorkspaceStoreOption {
 // NewWorkspaceStore returns a new WorkspaceStore.
 func NewWorkspaceStore(l logging.Logger, opts ...WorkspaceStoreOption) *WorkspaceStore {
 	ws := &WorkspaceStore{
-		store:  map[types.UID]*Workspace{},
-		logger: l,
-		mu:     sync.Mutex{},
-		fs:     afero.Afero{Fs: afero.NewOsFs()},
+		store:    map[types.UID]*Workspace{},
+		logger:   l,
+		mu:       sync.Mutex{},
+		fs:       afero.Afero{Fs: afero.NewOsFs()},
+		executor: exec.New(),
 	}
 	for _, f := range opts {
 		f(ws)
@@ -90,7 +91,9 @@ type WorkspaceStore struct {
 	logger logging.Logger
 
 	mu sync.Mutex
-	fs afero.Afero
+
+	fs       afero.Afero
+	executor exec.Interface
 }
 
 // Workspace makes sure the Terraform workspace for the given resource is ready
@@ -121,7 +124,7 @@ func (ws *WorkspaceStore) Workspace(ctx context.Context, c resource.SecretClient
 	ws.mu.Lock()
 	w, ok := ws.store[tr.GetUID()]
 	if !ok {
-		ws.store[tr.GetUID()] = NewWorkspace(dir, WithLogger(l))
+		ws.store[tr.GetUID()] = NewWorkspace(dir, WithLogger(l), WithExecutor(ws.executor))
 		w = ws.store[tr.GetUID()]
 	}
 	ws.mu.Unlock()
@@ -134,8 +137,8 @@ func (ws *WorkspaceStore) Workspace(ctx context.Context, c resource.SecretClient
 	if !os.IsNotExist(err) {
 		return w, nil
 	}
-	cmd := exec.CommandContext(ctx, "terraform", "init", "-input=false")
-	cmd.Dir = w.dir
+	cmd := w.executor.CommandContext(ctx, "terraform", "init", "-input=false")
+	cmd.SetDir(w.dir)
 	out, err := cmd.CombinedOutput()
 	l.Debug("init ended", "out", string(out))
 	return w, errors.Wrapf(err, "cannot init workspace: %s", string(out))
--- a/pkg/terraform/workspace.go
+++ b/pkg/terraform/workspace.go
@@ -19,12 +19,13 @@ package terraform
 import (
 	"context"
 	"os"
-	"os/exec"
 	"path/filepath"
 	"strings"
 	"time"
 
 	"github.com/pkg/errors"
+	"github.com/spf13/afero"
+	k8sExec "k8s.io/utils/exec"
 
 	"github.com/crossplane/crossplane-runtime/pkg/logging"
 
@@ -46,12 +47,35 @@ func WithLogger(l logging.Logger) WorkspaceOption {
 	}
 }
 
+// WithExecutor sets the executor of Workspace.
+func WithExecutor(e k8sExec.Interface) WorkspaceOption {
+	return func(w *Workspace) {
+		w.executor = e
+	}
+}
+
+// WithLastOperation sets the Last Operation of Workspace.
+func WithLastOperation(lo *Operation) WorkspaceOption {
+	return func(w *Workspace) {
+		w.LastOperation = lo
+	}
+}
+
+// WithAferoFs lets you set the fs of WorkspaceStore.
+func WithAferoFs(fs afero.Fs) WorkspaceOption {
+	return func(ws *Workspace) {
+		ws.fs = afero.Afero{Fs: fs}
+	}
+}
+
 // NewWorkspace returns a new Workspace object that operates in the given
 // directory.
 func NewWorkspace(dir string, opts ...WorkspaceOption) *Workspace {
 	w := &Workspace{
 		LastOperation: &Operation{},
 		dir:           dir,
+		logger:        logging.NewNopLogger(),
+		fs:            afero.Afero{Fs: afero.NewOsFs()},
 	}
 	for _, f := range opts {
 		f(w)
@@ -69,9 +93,12 @@ type Workspace struct {
 	// LastOperation contains information about the last operation performed.
 	LastOperation *Operation
 
-	dir    string
-	env    []string
-	logger logging.Logger
+	dir string
+	env []string
+
+	logger   logging.Logger
+	executor k8sExec.Interface
+	fs       afero.Afero
 }
 
 // ApplyAsync makes a terraform apply call without blocking and calls the given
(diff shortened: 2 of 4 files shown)


pkg/terraform/workspace_test.go
```diff
@@ -0,0 +1,369 @@
...
+	changeSummaryUpdate   = `{"@level":"info","@message":"Plan: 0 to add, 1 to change, 0 to destroy.","@module":"terraform.ui","@timestamp":"0000-00-00T00:00:00.000000+03:00","changes":{"add":0,"change":1,"remove":0,"operation":"plan"},"type":"change_summary"}`
+	changeSummaryNoAction = `{"@level":"info","@message":"Plan: 0 to add, 0 to change, 0 to destroy.","@module":"terraform.ui","@timestamp":"0000-00-00T00:00:00.000000+03:00","changes":{"add":0,"change":0,"remove":0,"operation":"plan"},"type":"change_summary"}`
+)
+
+func newLogger() logging.Logger {
+	zl := zap.New(zap.UseDevMode(true))
+	return logging.NewLogrLogger(zl.WithName("provider-jet-test"))
+}
+
+func TestWorkspace_Apply(t *testing.T) {
```
negz, line 52: Nit: Underscores aren't particularly idiomatic in go function names in my experience, so I'd prefer us to avoid them in test function names too. I think `TestWorkspaceApply` would be fine here, for example.

https://github.com/crossplane/crossplane/blob/master/CONTRIBUTING.md#prefer-table-driven-tests

sergenyalcin: Thank you @negz Resolved!

pkg/terraform/workspace_test.go
```diff
@@ -0,0 +1,454 @@
...
+	return NewWorkspace(directory, WithExec(&testingexec.FakeExec{
+		CommandScript: []testingexec.FakeCommandAction{
+			func(_ string, _ ...string) k8sExec.Cmd {
+				b := bytes.Buffer{}
+				b.WriteString(errBoom.Error())
+
+				return &testingexec.FakeCmd{
+					CombinedOutputScript: []testingexec.FakeAction{
+						func() ([]byte, []byte, error) {
+							return b.Bytes(), nil, errBoom
```
muvaf, line 69:
```suggestion
				return &testingexec.FakeCmd{
					CombinedOutputScript: []testingexec.FakeAction{
						func() ([]byte, []byte, error) {
							return []byte(errBoom.Error()), nil, errBoom
```

pkg/terraform/workspace_test.go
```diff
@@ -0,0 +1,454 @@
...
+						},
+					},
+				}
+			},
+		},
+	},
+	), WithLogger(newLogger()))
```
muvaf, line 76: In general, we initialize structs with `NewNopLogger()` in their `New*` function by default but it seems we don't do that for `NewWorkspace(..)` function. Would you mind adding that and removing these `WithLogger` statements from tests?

pkg/terraform/workspace_test.go
```diff
@@ -0,0 +1,454 @@
...
+)
+
+var (
+	testType              = "very-cool-type"
+	applyType             = "apply"
+	lineage               = "very-cool-lineage"
+	terraformVersion      = "1.0.10"
+	version               = 1
+	serial                = 3
+	directory             = "./testdata"
```
muvaf, line 44: We can use `afero`'s in-memory file system for test data, similar to the following test https://github.com/crossplane-contrib/terrajet/blob/8e96bd1/pkg/terraform/files_test.go#L89

muvaf: I noticed that `WorkspaceStore` abstracts the filesystem with `afero` but `Workspace` doesn't. How would you feel about making it use `afero` and use that in your tests?

pkg/terraform/workspace_test.go
```diff
@@ -0,0 +1,454 @@
...
+			},
+			want: want{
+				r: ApplyResult{
+					State: state,
+				},
+			},
+		},
+		"Failure": {
+			args: args{
+				w: newFailureWorkspace(),
```
muvaf, line 144: I wonder if we can come up with a function that is capable of being called for all cases in a table so that it's more consistent and easy to read. It'd be awesome if we'd be able to use existing `NewWorkspace` with function options though that could be a bit tricky with afero filesystem where you'd need a function context to fill the files in the memory.

muvaf: Another option could be to accept more granular things, like `executor`, filesystem and possible last operation and then use them to initialize `NewWorkspace` in `t.Run`

pkg/terraform/workspace_test.go
```diff
@@ -0,0 +1,454 @@
...
+func TestWorkspacePlan(t *testing.T) {
+	type args struct {
+		w *Workspace
+	}
+	type want struct {
+		r   PlanResult
+		err error
+	}
+
+	cases := map[string]struct {
```
muvaf, line 278: A failure case for `Plan` would be useful. I'd suggest running this file's tests with `Run with Coverage` feature of Goland or IDE you have so that it highlights which lines aren't covered.

pkg/terraform/store.go
```diff
@@ -28,6 +28,7 @@ import (
 	"github.com/pkg/errors"
 	"github.com/spf13/afero"
 	"k8s.io/apimachinery/pkg/types"
+	k8sExec "k8s.io/utils/exec"
```
muvaf, line 31:
```suggestion
	kexec "k8s.io/utils/exec"
```

We usually go for prefixing with `k` if there is a counterpart, like `kerrors` for k8s errors package.

muvaf: In fact, it seems there is one last usage of `os/exec` in this file in line 148 where we call `terraform init`, so if we remove that, we can just use `exec`.

pkg/terraform/workspace.go
```diff
@@ -46,6 +46,20 @@ func WithLogger(l logging.Logger) WorkspaceOption {
 	}
 }
 
+// WithExec sets the executor of Workspace.
+func WithExec(e k8sExec.Interface) WorkspaceOption {
```
muvaf, line 50: I think this could be renamed to `WithExecutor` or we can rename the field name `executor` to `exec`.

pkg/terraform/workspace.go
```diff
@@ -266,10 +282,9 @@ func (w *Workspace) Plan(ctx context.Context) (PlanResult, error) {
...
-func (w *Workspace) configureCmd(cmd *exec.Cmd) {
-	if cmd.Env == nil {
-		cmd.Env = os.Environ()
-	}
-	cmd.Env = append(cmd.Env, w.env...)
-	cmd.Dir = w.dir
+func (w *Workspace) configureCmd(cmd k8sExec.Cmd) {
+	env := os.Environ()
+	env = append(env, w.env...)
+	cmd.SetEnv(env)
```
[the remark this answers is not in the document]
muvaf: nit: I think moving these two functions to where `configureCmd` is called would be easier to follow for readers.

pkg/terraform/workspace_test.go
```diff
@@ -0,0 +1,454 @@
...
+
+func newLogger() logging.Logger {
+	zl := zap.New(zap.UseDevMode(true))
+	return logging.NewLogrLogger(zl.WithName("provider-jet-test"))
+}
+
+func timeNow() *time.Time {
+	t := time.Unix(0, 0)
+	return &t
+}
```
muvaf, line 108:
```suggestion
```

Having `var now = time.Now()` and referencing it via `&now` could be less code.

muvaf (commented):
Great start @sergenyalcin !

pkg/terraform/workspace.go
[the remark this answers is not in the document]
sergenyalcin: Moved these two functions and configureCmd function was removed.

pkg/terraform/workspace_test.go
sergenyalcin: In `NewWorkspace` function, `NewNopLogger` was added as a default

pkg/terraform/workspace_test.go
sergenyalcin: `afero` abstraction was added for `Workspace` and it was used for tests.

pkg/terraform/workspace_test.go
sergenyalcin: I tried to create `Workspace` objects by using the `NewWorkspace` and its `options`. For creating `FakeExec`, I have added a utility function and I passed this to the `WithExecutor`. For filling the `afero` file, in `t.Run`, `WriteFile` was used.

pkg/terraform/workspace_test.go
sergenyalcin: Added. Thanks for your suggestion!

pkg/terraform/store.go
sergenyalcin: Done!

pkg/terraform/workspace.go
sergenyalcin: The reason of this naming is that, in this package, I have added another function `WithExecutor` for `WorkspaceStore`. As you know, Go does not allow this redeclaration.

While looking your comments, I realized that, we do not use the `WithExecutor` for `WorkspaceStore` and I deleted this one and I renamed the `WithExec` to `WithExecutor`. However, I think this is not a good solution for this because in future, we can need to use `WithExecutor` for `WorkspaceStore` I wanted to discuss this issue in this thread.

Another example is that, I had to name the Afero option function for `Workspace` with `WithAferoFs` name. Beacuse WithFs and WithFileSystem names are already used.

I do not resolve this conversation. What are your thoughts?

pkg/terraform/workspace_test.go
sergenyalcin: Done!

sergenyalcin:
> Great start @sergenyalcin !

Thanks for your review @muvaf I addressed your comments.

pkg/terraform/workspace.go
muvaf: We can consider separating them into different packages but that's not really concern of this PR. So, it should be fine then.

pkg/terraform/store.go
```diff
@@ -91,6 +92,8 @@ type WorkspaceStore struct {
 
 	mu sync.Mutex
 	fs afero.Afero
+
+	executor exec.Interface
```
muvaf, line 96: nitpick
```suggestion

	fs afero.Afero
	executor exec.Interface
```

pkg/terraform/workspace.go
```diff
@@ -72,6 +96,10 @@ type Workspace struct {
 	dir    string
 	env    []string
 	logger logging.Logger
+
+	executor k8sExec.Interface
+
+	fs afero.Afero
```
muvaf, line 102: nitpick
```suggestion
	
	logger logging.Logger
	executor k8sExec.Interface
	fs afero.Afero
```

pkg/terraform/workspace_test.go
```diff
@@ -0,0 +1,441 @@
...
+				called: true,
+			},
+		},
+	}
+
+	for name, tc := range cases {
+		t.Run(name, func(t *testing.T) {
+			err := tc.w.ApplyAsync(tc.c)
+			if t.Name() == "TestWorkspaceApplyAsync/Callback" {
+				called := <-calls
```
muvaf, line 375: nice trick!

muvaf (approved):
Thank you @sergenyalcin !

pkg/terraform/store.go
sergenyalcin: Done!

pkg/terraform/workspace.go
sergenyalcin: Done!

PR #174, opened 2021-12-15
Conversion from tfjson to consume schema from cli output

### LINKED ISSUE #131: Consider getting provider schema using Terraform CLI

Today, we are generating CRD schemas by importing the package with the schema definition from Terraform provider codebase. While this worked well so far, we should consider breaking this dependency because:

- Terraform documentation explicitly mentions that importing Provider as Go Modules is an [unsupported way to interact](https://www.terraform.io/docs/extend/best-practices/depending-on-providers.html#do-not-import-providers-as-go-modules). 
- We only need this at code generation time and do not have any dependency at runtime. 
- Some Terraform providers have the schema under `internal` directory which prevents importing them as is (and needs some workarounds)

As suggested in the [Terraform docs](https://www.terraform.io/docs/extend/best-practices/depending-on-providers.html#exporting-the-schema), we can just use the Terraform CLI to fetch the schema of the Provider and break go module dependency.

> Some projects just care about the schema and resources a provider presents. As of Terraform 0.12, the `terraform providers schema -json` command can be used to export a JSON representation of the schemas for the providers used in a workspace.

### LINKED ISSUE #175: Generate provider schema only with CLI output

This is a follow-up issue for #131 where we have investigated the approach and came up with a [draft PR](https://github.com/crossplane/terrajet/pull/174). 

Work items to be covered as part of this PR:

- [x] Test #174 with existing jet providers and ensure it ends up with the same schema with expected exceptions => https://github.com/crossplane-contrib/provider-jet-gcp/pull/38
- [x] Find and implement an appropriate solution to `Number` -> `Int/Float`.
>   Seems like using `Float` for `Number` types is reasonable, since I noticed that `Float` type appears as `Number` in CRD schema. e.g. Number(tfjson) -> "float(terrajet)" -> number(CRD)
- [x] Automate getting schema from cli, e.g. with a Makefile target. This should probably land on provider-jet-template repo. => https://github.com/crossplane-contrib/provider-jet-template/pull/13
- [x] Update the provider generation Guide with the new approach.

turkenh:

Fixes #175 in the scope of #131 

I have:

See the provider-gcp schema generated using the conversion here: https://github.com/crossplane-contrib/provider-jet-gcp/pull/38/

Tested with provider-github with the updated jet template in this PR: https://github.com/crossplane-contrib/provider-jet-template/pull/13


pkg/types/conversion/sdkv1/sdkv1.go
```diff
@@ -14,7 +14,7 @@
  limitations under the License.
 */
 
-package conversion
+package sdkv1
```
muvaf, line 17: Can we mark the stuff in this package deprecated?

pkg/types/conversion/tfjson/tfjson.go
```diff
@@ -0,0 +1,221 @@
...
+// representation of resource schema map, e.g. output of `terraform providers schema -json`
+// to plugin sdk representation. This is mostly the opposite of what the
+// following method is doing: https://github.com/hashicorp/terraform-plugin-sdk/blob/7e0a333644f1971a936995677b7a106140a0659f/helper/schema/core_schema.go#L43
+//
+// Ideally, we should not rely on plugin SDK types in Terrajet at all but only
+// work with types in https://github.com/hashicorp/terraform-json which is
+// there exactly for this purpose, an external representation of Terraform
+// schemas. This conversion aims to be an intermediate step for that ultimate
+// goal.
+func GetV2ResourceMap(resourceSchemas map[string]*tfjson.Schema) map[string]*schemav2.Resource {
```
muvaf, line 40: I wonder how we can help provider builders to know that the old way is deprecated and that's how we'll do this going forward. `tjconfig.NewProvider` still accepts a resource map but the end goal for this work is to make it accept the CLI schema, maybe even as a `string`. Something like `tjconfig.NewProviderWithSchema` and deprecate `tjconfig.NewProvider`?

pkg/types/conversion/tfjson/tfjson.go
```diff
@@ -0,0 +1,221 @@
...
+)
+
+// GetV2ResourceMap converts input resource schemas with
+// "terraform-json" representation to terraform-plugin-sdk representation which
+// is what Terrajet expects today.
+//
+// What we are trying to achieve here is to convert a lower level
+// representation of resource schema map, e.g. output of `terraform providers schema -json`
+// to plugin sdk representation. This is mostly the opposite of what the
+// following method is doing: https://github.com/hashicorp/terraform-plugin-sdk/blob/7e0a333644f1971a936995677b7a106140a0659f/helper/schema/core_schema.go#L43
```
muvaf, line 33: I'm not sure how long we'd like to keep the backward compatibility, especially if there is no CRD API change, but the next step could be to make builder work with CLI schema directly and do the `sdk -> cli schema` conversion using this referenced tool. You may consider doing that in this PR, too, if the conversions we have here are too similar to that file anyway.

muvaf (commented):
Thanks @turkenh ! Do you have data about whether we see CRD changes with this in bigger providers, like jet-aws or jet-gcp with full set?

turkenh:
> Thanks @turkenh ! Do you have data about whether we see CRD changes with this in bigger providers, like jet-aws or jet-gcp with full set?

@muvaf, not the full set but there is [this PR](https://github.com/crossplane-contrib/provider-jet-gcp/pull/38) generated with the default resource set in provider-jet-gcp.

Basically, there are two types of schema changes that I observed:
- Int64 to Float64 changes in API types and  `format: int64, type: integer` to `type: number` in corresponding CRD schema
- Some nested resources lose some information like description or [required](https://github.com/crossplane-contrib/provider-jet-gcp/pull/38/files#diff-f7411156edaa9a050e465c9522e70b19a2d1261643a892a7fe9ec24db981d42aR739) because of a more limited representation. 

I haven't dug further since we first want to experiment with new providers first.

pkg/types/conversion/tfjson/tfjson.go
muvaf: With `tjconfig.NewProviderWithSchema(schema string, ...` calling `GetV2ResourceMap` and the rest would be same as `tjconfig.NewProvider`.

pkg/types/conversion/tfjson/tfjson.go
turkenh: As mentioned in the next comment lines, I agree that ultimately our builder should work with cli schema. However, I still see value to experiment current approach with new providers for a while.

I liked the idea of working with CLI schema in type builder but doing the conversion from `sdk -> cli schema`, but I wouldn't prefer to do in this PR because:

- Making type builder working with CLI would require a good amount of work and I would not prefer to delay this feature for new providers since there are some people struggling with the current approach, i.e. importing tf provider go package.
- Since the `cli schema` is a more limited representation than the `sdk` one, we would still lose some information, e.g. int/float distinction or lack of required attribute for a deeply nested object field, etc. So, at this point (having one conversion already implemented), I don't see much benefit of spending time the other way. 

Stating this, I believe we can prioritize to work on a follow-up PR updating type builder to work with cli schema and speed up the experimenting process.

pkg/types/conversion/tfjson/tfjson.go
turkenh: That's a great point @muvaf!
Done!

turkenh:
@muvaf I've resolved/responded to your comments and this is now ready for another pass.

pkg/config/provider.go
```diff
@@ -143,7 +146,28 @@ func WithDefaultResourceFn(f DefaultResourceFn) ProviderOption {
 	}
 }
 
+// NewProviderWithSchema builds and returns a new Provider from provider
+// tfjson schema, that is generated using Terraform CLI with:
+// `terraform providers schema --json`
+func NewProviderWithSchema(schema string, prefix string, modulePath string, opts ...ProviderOption) *Provider {
```
muvaf, line 152:
```suggestion
func NewProviderWithSchema(schema []byte, prefix string, modulePath string, opts ...ProviderOption) *Provider {
```

muvaf (approved):
Thanks @turkenh ! LGTM!

I think the only risk with this change is the usage of `float64` which upstream folks have [some concerns](https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api-conventions.md#primitive-types) about it. Though I'm not sure how `IntOrString` type would work in our runtime tools. Maybe it's no different than others since we use fieldpath?

Another thing regarding float is how Terraform works with it, i.e. if we print `0.0` instead of `0` in `main.tf.json`, would TF CLI and TF provider be OK with that?

turkenh:
Ok, I did some testing and writing here for future reference.

I've selected google [google_compute_autoscaler](https://registry.terraform.io/providers/hashicorp/google/latest/docs/resources/compute_autoscaler) resource and it [autoscaling_policy.cpu_utilization.target ](https://registry.terraform.io/providers/hashicorp/google/latest/docs/resources/compute_autoscaler#target) as it expects a float value.

**Code Generation:**

Here is how it looks like in provider schema tf output:

```json
"target": {
    "type": "number",
    "description": "The target CPU utilization that the autoscaler should maintain.\nMust be a float value in the range (0, 1]. If not specified, the\ndefault is 0.6.\n\nIf the CPU level is below the target utilization, the autoscaler\nscales down the number of instances until it reaches the minimum\nnumber of instances you specified or until the average CPU of\nyour instances reaches the target utilization.\n\nIf the average CPU is above the target utilization, the autoscaler\nscales up until it reaches the maximum number of instances you\nspecified or until the average utilization reaches the target\nutilization.",
    "description_kind": "plain",
    "required": true
}
```

In the generated schema:

```go
Target *float64 `json:"target" tf:"target,omitempty"`
```

In CRD Spec:

```
target:
  description: "The target CPU utilization ...."
  type: number
```

**Runtime:**

Created following CR:

```
apiVersion: compute.gcp.jet.crossplane.io/v1alpha1
kind: Autoscaler
metadata:
  name: numbertest
spec:
  forProvider:
    zone: us-central1-a
    target: projects/<redacted>/zones/us-central1-a/instanceGroupManagers/hasan-another
    autoscalingPolicy:
      - minReplicas: 1
        maxReplicas: 4
        cooldownPeriod: 59
        cpuUtilization:
          - target: 0.43
```

Generated `main.tf` by Terrajet:

```
{
  "resource": {
    "google_compute_autoscaler": {
      "numbertest": {
        "autoscaling_policy": [
          {
            "cooldown_period": 59,
            "cpu_utilization": [
              {
                "target": 0.43
              }
            ],
            "max_replicas": 4,
            "min_replicas": 1
          }
        ]
        ...
      }
    }
  },
}
```

Resource successfully created and reported Synced & Ready as `true`. And here is how it looks like in GCP console:

Screen Shot 2022-02-21 at 13 12 41

Then I updated target to 1 as below:

```
    autoscalingPolicy:
        ...
        cpuUtilization:
          - target: 1
```

With the generated `main.tf`:

```
            "cpu_utilization": [
              {
                "target": 1
              }
            ],
```

It updated and synced as below:

Screen Shot 2022-02-21 at 13 14 43

turkenh:
> Thanks @turkenh ! LGTM!
> 
> I think the only risk with this change is the usage of `float64` which upstream folks have [some concerns](https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api-conventions.md#primitive-types) about it. Though I'm not sure how `IntOrString` type would work in our runtime tools. Maybe it's no different than others since we use fieldpath?
> 
> Another thing regarding float is how Terraform works with it, i.e. if we print `0.0` instead of `0` in `main.tf.json`, would TF CLI and TF provider be OK with that?

@muvaf in the above experiment, I've validated that we can work with [numbers](https://www.terraform.io/language/expressions/types#number) and can both provide integers and/or floats for such fields. With that, I am feeling good with numbers.

muvaf:
@turkenh Thanks for that extensive test! I think this could warrant a Terrajet release.

muvaf:
Actually let's get https://github.com/crossplane/terrajet/pull/195 as well included in the release.

PR #179, opened 2021-12-16
Capitalize prefix/suffix acronyms in default resource config's kind name

### LINKED ISSUE #178: Capitalize Acronyms in Generated Resource Kind Names

We would like to have the acronyms appearing as prefixes or suffixes in generated resource kind names to be capitalized. For example, instead of `Serverid`, we would like to have the default kind name as `ServerID`. 

We could capitalize prefix/suffix acronyms appearing in the kind name of the default configuration.

ulucinar:

Fixes #178 

In the default resource config, capitalize acronyms appearing as prefix/suffixes in kind names. For example, instead of `Serverid`, we would like to have the default kind name as `ServerID`. 

I have:

With `provider-jet-azure` by generating resources from main branch.


pkg/config/defaults.go
```diff
@@ -95,7 +97,7 @@ func DefaultResource(name string, terraformSchema *schema.Resource, opts ...Reso
 		Name:              name,
 		TerraformResource: terraformSchema,
 		ShortGroup:        group,
-		Kind:              kind,
+		Kind:              typeName.CapitalizeAcronyms(kind),
```
muvaf, line 100:
```suggestion
		Kind:              types.NewNameFromCamel(kind).Camel,
```
or
```suggestion
                // kind could be initialized with NewNameFromSnake
		Kind:              kind.Camel,
```

Is there an advantage over re-using the `Name` type? From the look of it, it should produce the same output.

ulucinar: I've pushed another commit that assumes acronyms are always tokens separated by underscores. The previous implementation does not make that assumption, i.e., whether an acronym is a full token or it's a prefix or suffix of a token, it's processed. Thus, the previous implementation is more aggressive and assuming/processing acronyms only if they are full tokens is more conservative. 

I think we can, by default, generate more conservative names.

pkg/types/reference.go
```diff
@@ -42,8 +43,8 @@ func (g *Builder) generateReferenceFields(t *types.TypeName, f *types.Var, r con
 		sfn = f.Name() + "Selector"
 	}
 
-	rn := NewNameFromCamel(rfn)
-	sn := NewNameFromCamel(sfn)
+	rn := name.NewNameFromCamel(rfn)
```
turkenh, line 46: Do we really need to move those into a new package? If yes, may be we should consider removing package name from the methods, e.g. `name.NewNameFromCamel` => name.NewFromCamel

pkg/config/defaults.go
ulucinar: I had to move the name package to get rid of an import cycle between the config and types packages. Done the renaming.

pkg/types/reference.go
ulucinar: I had to move the name package to get rid of an import cycle between the config and types packages. Done the renaming.

turkenh (approved):
LGTM, thanks @ulucinar !

PR #182, opened 2021-12-20
Assume string type for unset Set/List/Map element types

### LINKED ISSUE #177: CRD pipeline fails if a Terraform Schema of a Set/List/Map type does not set its Element Type
### What happened?

For `terraform-provider-mongodbatlas`, we cannot generate types for `mongodbatlas_cluster`, `mongodbatlas_advanced_cluster`, `mongodbatlas_encryption_at_rest`, `mongodbatlas_global_cluster_config`, `mongodbatlas_cloud_provider_access` because they contain Schema definitions of map types which do not set their element types. [Here](https://github.com/crossplane-contrib/terrajet/blob/2bfe49174387c95b7f5c510290cd8eed1d726641/pkg/types/builder.go#L240), we check the element type of the Map/List/Set type and if it's not set, i.e., `nil`, we error [here](https://github.com/crossplane-contrib/terrajet/blob/2bfe49174387c95b7f5c510290cd8eed1d726641/pkg/types/builder.go#L283). 

If you look at the following code comments for element types:
```golang
	// Elem represents the element type. For a TypeMap, it must be a *Schema
	// with a Type that is one of the primitives: TypeString, TypeBool,
	// TypeInt, or TypeFloat. Otherwise it may be either a *Schema or a
	// *Resource. If it is *Schema, the element type is just a simple value.
	// If it is *Resource, the element type is a complex structure,
	// potentially managed via its own CRUD actions on the API.
```
it sounds like the elem type _must_ be set. We may:

1. Have it fixed in the corresponding Terraform provider repo
2. Ignore generation of the resource (not a real solution)
3. Modify Terraform resource schema while configuring the Terraformed resource (this can be quite ugly)
4. If the field is deprecated, maybe remove (delete) it from the Terraform schema (similar ugliness factor with the above solution)
5. Assume the type to be of string
6. ...

I think it's worth taking a look at the 5th option above. Especially assuming not setting the elem type for a map type actually works for Terraform, why should we fail? I'd guess that terraform handles such cases with a default type, as a string?

Another case for the 5th option is the `terraform providers schema` displays the elem type as `string` for an unset elem type:
```json
"connection_strings": {
                "type": [
                  "list",
                  [
                    "object",
                    {
                      "aws_private_link": [
                        "map",
                        "string"
                      ],
```

Another thing to note here is that the message we return `element type of ... should be either schema.Resource or schema.Schema` is quite misleading here. It can be one of `ValueType` (representing a primitive type), `Schema` or `Resource`. For maps, only primitives are allowed per documentation.

### How can we reproduce it?

Try to generate a provider for `terraform-provider-mongodbatlas` `v1.1.1`.

ulucinar:

Fixes #177 

Assume that an unset Terraform List/Set/Map element schema is of string type. Please refer to #177 for more context.

I have:

Tested with https://github.com/crossplane-contrib/provider-jet-mongodbatlas

### DIFF
--- a/pkg/types/builder.go
+++ b/pkg/types/builder.go
@@ -283,6 +283,10 @@ func (g *Builder) buildSchema(sch *schema.Schema, cfg *config.Resource, tfPath [
 				}
 				elemType = paramType
 			}
+		// if unset
+		// see: https://github.com/crossplane/terrajet/issues/177
+		case nil:
+			elemType = types.Universe.Lookup("string").Type()
 		default:
 			return nil, errors.Errorf("element type of %s should be either schema.Resource or schema.Schema", fieldPath(names))
 		}


pkg/types/builder.go
```diff
@@ -280,6 +280,9 @@ func (g *Builder) buildSchema(sch *schema.Schema, cfg *config.Resource, tfPath [
 				}
 				elemType = paramType
 			}
+		// if unset
```
muvaf, line 283: nitpick: A more explanatory comment or link to the issue would be more helpful for future readers.

muvaf (approved):

PR #183, opened 2021-12-21
New acronyms

muvaf:

Adds new acronyms that I needed in Jet AWS to the list. Some of them are AWS-specific, like EC2, but I think that's probably harmless since there is no context that they can be real words.

I have:

With Jet AWS.


pkg/config/defaults.go
```diff
@@ -87,17 +86,17 @@ func DefaultResource(name string, terraformSchema *schema.Resource, opts ...Reso
 	// - aws_rds_cluster => Cluster
 	// - aws_rds_cluster_parameter_group => ClusterParameterGroup
 	// - kafka_topic => Topic
-	kind := strcase.ToCamel(strings.Join(words[2:], "_"))
+	kind := tjname.NewFromSnake(strings.Join(words[2:], "_")).Camel
```
muvaf, line 89: Otherwise, `strcase.ToCamel` treates `ec2` as two words `Ec` and `2`, which causes it not to be found in the acronyms list.

turkenh: I needed this in provider-jet-gcp as well for hl7 to HL7 👍

turkenh (approved):

PR #187, opened 2021-12-29
Update Configuring a Resource guide

### LINKED ISSUE #153: Update configuring resource guide per latest changes
### What happened?

There are couple of changes in Terrajet configuration interface that happened after the `Configuring a resource` guide is introduced [first time](https://github.com/crossplane-contrib/provider-jet-aws/pull/123).

We need to update the guide accordingly.

There are a couple of review comments in https://github.com/crossplane-contrib/terrajet/pull/130 which we would need to tackle first: 

- https://github.com/crossplane-contrib/terrajet/pull/130#discussion_r749507600
- https://github.com/crossplane-contrib/terrajet/pull/130#discussion_r749508451
- https://github.com/crossplane-contrib/terrajet/pull/130#discussion_r752947141
- https://github.com/crossplane-contrib/terrajet/pull/130#discussion_r752947828
- https://github.com/crossplane-contrib/terrajet/pull/130#discussion_r749516661

### How can we reproduce it?

Reflect latest changes to the guide.

turkenh:

This PR updates Configuring a Resource guide:

- For latest changes in External name configuration
- Adds a section for overriding Terraform schema

Fixes #153 

I have:

N/A


docs/configuring-a-resource.md
```diff
@@ -22,126 +22,239 @@ cloud provider; sometimes it could simply be the name of resource
...
+  // name and assign it to that specific key for that resource type.
+  SetIdentifierArgumentFn SetIdentifierArgumentsFn
+  
+  // GetExternalNameFn returns the external name extracted from TF State. In most cases,
+  // "id" field contains all the information you need. You'll need to extract
+  // the format that is decided for external name annotation to use.
+  // For example the following is an Azure resource ID:
+  // /subscriptions/00000000-0000-0000-0000-000000000000/resourceGroups/mygroup1
+  // The function should return "mygroup1" so that it can be used to set external
+  // name if it was not set already.
```
displague, line 63: Perhaps you could say why this is needed. Why can't the external name be `/subscriptions/00000000-0000-0000-0000-000000000000/resourceGroups/mygroup1`?

docs/configuring-a-resource.md
```diff
@@ -22,126 +22,239 @@ cloud provider; sometimes it could simply be the name of resource
...
-    // that sets external name to metadata.name if none specified should be disabled.
-    // It needs to be disabled for resources whose external name includes information
-    // more than the actual name of the resource, like subscription ID or region
-    // etc. which is unlikely to be included in metadata.name
-    DisableNameInitializer bool
+  // SetIdentifierArgumentFn sets the name of the resource in Terraform argument
+  // map. In many cases, there is a field called "name" in the HCL schema, however,
+  // there are cases like RDS DB Cluster where the name field in HCL is called
+  // "cluster_identifier". This function is the place that you can take external
+  // name and assign it to that specific key for that resource type.
```
displague, line 54: It would be helpful to explain, perhaps in the `ExternalName` comment block, what the distinguishing features of 'name' and 'identifier' are. When can these be the same, when should they differ. Finding that `SetIdentifierArgumentFn sets the name` confuses me because `identifier` and `name` seem to be interchangeable.

docs/configuring-a-resource.md
```diff
@@ -22,126 +22,239 @@ cloud provider; sometimes it could simply be the name of resource
...
+  // the format that is decided for external name annotation to use.
+  // For example the following is an Azure resource ID:
+  // /subscriptions/00000000-0000-0000-0000-000000000000/resourceGroups/mygroup1
+  // The function should return "mygroup1" so that it can be used to set external
+  // name if it was not set already.
+  GetExternalNameFn GetExternalNameFn
+  
+  // GetIDFn returns the string that will be used as "id" key in TF state. In
+  // many cases, external name format is the same as "id" but when it is not
+  // we may need information from other places to construct it. For example,
```
displague, line 68: This adds to the equivocation that I mentioned in https://github.com/crossplane/terrajet/pull/187/files#r777662083

Perhaps a section of the docs needs to be referenced that can go into more detail on the distinctions.

displague: I missed the example cases describing these scenarios later in the doc.  

It may be clearer to discuss the external name vs name vs IDs topic first before introducing these go fields.  

After that, the comments in these go fields may still be confusing. Perhaps the type can be offered in these docs without the comments (show the simple struct and link to the source for details). The text and examples in this documentation can describe how the fields can be used, and expand on usage within the scenarios and examples offered.

docs/configuring-a-resource.md
```diff
@@ -22,126 +22,239 @@ cloud provider; sometimes it could simply be the name of resource
 (e.g. VPC id ). This is something specific to resource, and we need some input
 configuration for terrajet to appropriately generate a resource.
 
-Since Terraform already needs the same identifier to import a resource, most
+Since Terraform already needs a similar [identifier to import a resource], most
```
muvaf, line 25: broken link

docs/configuring-a-resource.md
```diff
@@ -22,126 +22,239 @@ cloud provider; sometimes it could simply be the name of resource
...
+    }
 ```
 
-Please note, you can always check resource configurations of existing Providers
-as further examples under `config/<group>/config.go`.
+With this, we have covered most common scenarios for configuring external name.
+You can always check resource configurations of existing jet Providers as
+further examples under `config/<group>/config.go` in their repositories.
 
 ### Cross Resource Referencing
```
muvaf, line 260: In late-initialization behavior section, can we first describe the case where you'd need to customize and only then go into detail so that people can skip it in the beginning? In fact, we can have another high level section for configs like this one that should be done with data gathered from running the manual tests.

docs/configuring-a-resource.md
```diff
@@ -376,6 +489,50 @@ message in its `status.conditions`, we do the `LateInitializer.IgnoreFields`
...
+resource schema just works as is. However, there could be some rare edge cases
+like:
+
+- Field contains sensitive information but not marked as `Sensitive` or vice
+versa.
+- An attribute does not make sense to have in CRD schema, like
+[tags_all for jet AWS resources].
+- Field is of a collection type but [Elem] not set properly, e.g.
+[boot_disk.initialize_params.labels] in `google_compute_instance` (which would
+no longer be needed at least to default as `string` after [this PR]).
```
muvaf, line 518:
```suggestion
[boot_disk.initialize_params.labels] in `google_compute_instance`.
```

Do we have an example where you may want to change the given type?

docs/configuring-a-resource.md
```diff
@@ -376,6 +489,50 @@ message in its `status.conditions`, we do the `LateInitializer.IgnoreFields`
...
+  - Not Optional & Not Computed => Spec (required)
+  - Optional & Not Computed => Spec (optional)
+  - Optional & Computed => Spec (optional, to be late-initialized)
+  - Not Optional & Computed => Status
+
+Usually, we don't need to make any modifications in the resource schema and
+resource schema just works as is. However, there could be some rare edge cases
+like:
+
+- Field contains sensitive information but not marked as `Sensitive` or vice
```
muvaf, line 512: Addition of `region` field is also another example.

docs/configuring-a-resource.md
```diff
@@ -22,126 +22,239 @@ cloud provider; sometimes it could simply be the name of resource
...
-Since Terraform already needs the same identifier to import a resource, most
+Since Terraform already needs a similar [identifier to import a resource], most
 helpful part of resource documentation is the [import section].
 
-This is [the struct that holds the External Name configuration]:
+Terrajet performs some back and forth conversions between Crossplane resource
+model and Terraform configuration. We need a custom, per resource configuration
+to adapt Crossplane `external name` from Terraform `id`.
+
+![external name configuration](images/terrajet-externalname.png)
```
muvaf, line 32: nitpick: my two cents, even though the information on this diagram is needed only for the third case, it makes the whole thing look a bit more complicated if you never need the case 3, which I doubt will happen frequently in non-big three providers. Moving to under case 3 or removing would be my suggestion.

muvaf (approved):
Looks awesome! LGTM after the comments are addressed. Thanks!

docs/configuring-a-resource.md
turkenh: While I appreciate the curiosity behind this question, I don't think it is something that worths mentioning in this guide. It answers the question "why did we do this?" rather than "how can I configure my resource?" and better fits a design document rather than a guide. 

Stating  this, please check [this comment](https://github.com/crossplane/terrajet/pull/132#issuecomment-966663178) to learn the reasoning behind.

docs/configuring-a-resource.md
turkenh: Added as yet another example.

docs/configuring-a-resource.md
turkenh: With that PR merged, I am not aware of any example. Hence I removed this example.

docs/configuring-a-resource.md
turkenh: I remember that I had hard times to grasp why we need 3 different functions just to configure external name and visualizing it helps a lot not to got lost which is setting what. So, I see value with having this available somewhere.

However, I totally agree with your point that it is not a good initial experience. So, I will hide it as much as I can but make it still accessible from this document.

docs/configuring-a-resource.md
turkenh: Yeah, I agree that it is confusing with such similar words. 

May be it should be named as `SetNamingArgumentFn` to imply that it is setting the argument that gives the name of the resource?
If that sounds better, may be can consider renaming it (without breaking API of course by marking this one as deprecated and adding the other).

//cc @muvaf

PR #191, opened 2022-01-05
Configurable operation timeouts

### LINKED ISSUE #176: Ability to configure operation timeouts

Some terraform resources allows configuring [operation timeouts](https://www.terraform.io/docs/language/resources/syntax.html#operation-timeouts) for Read/Create/Update/Delete. There could be [some cases](https://github.com/crossplane-contrib/provider-jet-gcp/issues/12#issuecomment-992773746) that we want to configure these timeouts further at provider development time or at runtime via CR spec.

Enable configuring operation timeouts in Terrajet resource config to override during provider development.

turkenh:

Fixes #176 

I have:

- unit tests
- with provider gcp (WIP)

### DIFF
--- a/pkg/config/resource.go
+++ b/pkg/config/resource.go
@@ -18,6 +18,7 @@ package config
 
 import (
 	"context"
+	"time"
 
 	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
 	"github.com/pkg/errors"
@@ -184,6 +185,21 @@ func (s *Sensitive) AddFieldPath(tf, xp string) {
 	s.fieldPaths[tf] = xp
 }
 
+// OperationTimeouts allows configuring resource operation timeouts:
+// https://www.terraform.io/language/resources/syntax#operation-timeouts
+// Please note that, not all resources support configuring timeouts.
+type OperationTimeouts struct {
+	Read   *time.Duration
+	Create *time.Duration
+	Update *time.Duration
+	Delete *time.Duration
+}
+
+// TimeDuration returns a pointer to a given time.Duration.
+func TimeDuration(d time.Duration) *time.Duration {
+	return &d
+}
+
 // Resource is the set of information that you can override at different steps
 // of the code generation pipeline.
 type Resource struct {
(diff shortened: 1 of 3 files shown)


pkg/terraform/files.go
```diff
@@ -150,6 +150,30 @@ func (fp *FileProducer) WriteMainTF() error {
...
+	}
+	if fp.Config.OperationTimeouts.Update != nil {
+		timeouts["update"] = fp.Config.OperationTimeouts.Update.String()
+		timeoutsConfigured = true
+	}
+	if fp.Config.OperationTimeouts.Delete != nil {
+		timeouts["delete"] = fp.Config.OperationTimeouts.Delete.String()
+		timeoutsConfigured = true
+	}
+	if timeoutsConfigured {
```
muvaf, line 173:
```suggestion
	if len(timeouts) != 0 {
```

pkg/config/resource.go
```diff
@@ -184,6 +185,21 @@ func (s *Sensitive) AddFieldPath(tf, xp string) {
...
 }
 
+// OperationTimeouts allows configuring resource operation timeouts:
+// https://www.terraform.io/language/resources/syntax#operation-timeouts
+// Please note that, not all resources support configuring timeouts.
+type OperationTimeouts struct {
+	Read   *time.Duration
+	Create *time.Duration
+	Update *time.Duration
+	Delete *time.Duration
```
muvaf, line 195:
```suggestion
	Read   time.Duration
	Create time.Duration
	Update time.Duration
	Delete time.Duration
```

I am guessing that you had to use pointers to check whether it's given or not but there is a function `(t Time) IsZero() bool` we can use to do that. If that's the only reason, we could get rid of `TimeDuration(d time.Duration) *time.Duration` and work with value types.

turkenh: Yes, that was the reason and I was thinking that setting timeout to 0 could be a valid scenario, e.g. no timeout. But digging the terraform sdk code, I noticed that there is no such an implementation and probably no timeout does not make sense at all.

So, I would revert pointers but to check zero value I can not use `IsZero()` because it is on `time.Time` not `time.Duration`. Am I missing something?

Otherwise, I will go with the following:
```
	if t := fp.Config.OperationTimeouts.Read.String(); t != "0s" {
		timeouts["read"] = t
	}
```

muvaf: Oh yes, I missed that `IsZero` wouldn't work. I think you can check like `fp.Config.OperationTimeouts.Read != 0` since `type Duration int64`

muvaf (approved):
LGTM after it's tested manually. Thanks!

PR #197, opened 2022-01-09
Support []string and []*string as sensitive fields

### LINKED ISSUE #100: Support []string and []*string as sensitive fields

Tring to generate a CRD with sensitive field type []*string fails with:

```
panic: cannot generate crd: cannot build types for User: cannot build the types: got type "[]*string" for field "Passwords", only types "string" and "*string" supported as sensitive
```

```
config.Store.SetForResource("aws_elasticache_user", config.Resource{
		ExternalName: config.ExternalName{
			DisableNameInitializer: true,
		},
		Sensitive: config.Sensitive{
			CustomFieldPaths: []string{
				"passwords", // interestingly this field is not marked as sensitive in tf schema
			},
		},
		UseAsync: true,
	})
```

Support []string and []*string as sensitive fields

sergenyalcin:

Fixes #100 

This PR provides support for []string and []*string as sensitive fields. In resource manifests, more than one secrets and more than one fields of secrets can be referred as input for sensitive values. Also, sensitive field outputs will be written to the connection secret.

I have:

I have tested this PR by using the example in the issue description.
- Firstly, I added a custom configuration for `aws_elasticache_user` that marks the `passwords` field as sensitive.
- Then I created the following manifest:

```
apiVersion: elasticache.aws.jet.crossplane.io/v1alpha1
kind: User
metadata:
  name: sample-user
spec:
  forProvider:
    userName: "testUserName"
    accessString: "on ~app::* -@all +@read +@hash +@bitmap +@geo -setbit -bitfield -hset -hsetnx -hmset -hincrby -hincrbyfloat -hdel -bitop -geoadd -georadius -georadiusbymember"
    engine: "REDIS"
    region: us-west-1
    passwordsSecretRef:
      - name: test-slice-1
        namespace: crossplane-system
        key: pwd
      - name: test-slice-2
        namespace: crossplane-system
        key: pwd
  writeConnectionSecretToRef:
    name: test-conn
    namespace: default
```

- The external resource was successfully created and the connection secret had the correct values.

### DIFF
--- a/pkg/resource/sensitive.go
+++ b/pkg/resource/sensitive.go
@@ -36,6 +36,8 @@ const (
 	errCannotExpandWildcards          = "cannot expand wildcards"
 	errFmtCannotGetValueForFieldPath  = "cannot not get a value for fieldpath %q"
 	errFmtCannotGetStringForFieldPath = "cannot not get a string for fieldpath %q"
+	errFmtCannotGetSecretKeySelector  = "cannot get SecretKeySelector from xp resource for fieldpath %q"
+	errFmtCannotGetSecretValue        = "cannot get secret value for %v"
 )
 
 const (
@@ -44,6 +46,8 @@ const (
 	// overridden by any custom connection key configured which would break
 	// our ability to build tfstate back.
 	prefixAttribute = "attribute."
+
+	pluralSuffix = "s"
 
 	errGetAdditionalConnectionDetails = "cannot get additional connection details"
 	errFmtCannotOverrideExistingKey   = "overriding a reserved connection key (%q) is not allowed"
@@ -102,7 +106,7 @@ func GetConnectionDetails(attr map[string]interface{}, tr Terraformed, cfg *conf
 // GetSensitiveAttributes returns strings matching provided field paths in the
 // input data.
 // See the unit tests for examples.
-func GetSensitiveAttributes(from map[string]interface{}, mapping map[string]string) (map[string][]byte, error) {
+func GetSensitiveAttributes(from map[string]interface{}, mapping map[string]string) (map[string][]byte, error) { //nolint: gocyclo
 	if len(mapping) == 0 {
 		return nil, nil
 	}
@@ -124,10 +128,7 @@ func GetSensitiveAttributes(from map[string]interface{}, mapping map[string]stri
 			if v == nil {
 				continue
 			}
-			s, ok := v.(string)
-			if !ok {
-				return nil, errors.Errorf(errFmtCannotGetStringForFieldPath, fp)
-			}
+
 			// Note(turkenh): k8s secrets uses a strict regex to validate secret
 			// keys which does not allow having brackets inside. So, we need to
 			// do a conversion to be able to store as connection secret keys.
@@ -140,7 +141,21 @@ func GetSensitiveAttributes(from map[string]interface{}, mapping map[string]stri
 			if vals == nil {
 				vals = map[string][]byte{}
 			}
-			vals[fmt.Sprintf("%s%s", prefixAttribute, k)] = []byte(s)
+			switch s := v.(type) {
+			case []interface{}:
+				for i, e := range s {
+					value, ok := e.(string)
+					if !ok {
+						return nil, errors.Errorf(errFmtCannotGetStringForFieldPath, fp)
+					}
+					k = strings.TrimSuffix(k, pluralSuffix)
+					vals[fmt.Sprintf("%s%s.%v", prefixAttribute, k, i)] = []byte(value)
+				}
+			case string:
+				vals[fmt.Sprintf("%s%s", prefixAttribute, k)] = []byte(s)
+			default:
+				return nil, errors.Errorf(errFmtCannotGetStringForFieldPath, fp)
+			}
 		}
 	}
 	return vals, nil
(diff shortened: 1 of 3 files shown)


pkg/resource/sensitive.go
```diff
@@ -181,27 +202,58 @@ func GetSensitiveParameters(ctx context.Context, client SecretClient, from runti
...
+					return errors.Wrapf(err, "cannot set string as terraform attribute for fieldpath %q", tfPath)
+				}
 			}
 		}
 	}
 
 	return nil
 }
 
+func isSlice(v interface{}) bool {
```
ulucinar, line 249: We may replace this function with a type switch as you suggested.

sergenyalcin: Done! Using type switch will provide easy maintenance for new sensitive field types we plan to add in the future.

pkg/resource/sensitive.go
```diff
@@ -124,23 +126,37 @@ func GetSensitiveAttributes(from map[string]interface{}, mapping map[string]stri
...
+					// more details.
+					k, err := fieldPathToSecretKey(fp)
+					if err != nil {
+						return nil, errors.Wrapf(err, "cannot convert fieldpath %q to secret key", fp)
+					}
+					k = strings.TrimSuffix(k, pluralSuffix)
+					if vals == nil {
+						vals = map[string][]byte{}
+					}
+					vals[fmt.Sprintf("%s%s.%d", prefixAttribute, k, i)] = []byte(fmt.Sprintf("%v", e))
```
turkenh, line 146: Is there a specific reason to use `fmt.Sprintf("%v", e)` instead of a [type assertion](https://go.dev/ref/spec#Type_assertions) to string? 

I don't have a strong opinion but a little concerned because this would not be consistent with how we are handling single element, e.g. [fail](https://github.com/crossplane/terrajet/pull/197/files?diff=unified&w=0#diff-aee63bee02430ee5f3effa8c4f524ebc94fea5e2b200b5b4b53187ed1bc440fdR158) if not a string.

pkg/resource/sensitive.go
```diff
@@ -124,23 +126,37 @@ func GetSensitiveAttributes(from map[string]interface{}, mapping map[string]stri
...
+				}
+			case string:
+				k, err := fieldPathToSecretKey(fp)
+				if err != nil {
+					return nil, errors.Wrapf(err, "cannot convert fieldpath %q to secret key", fp)
+				}
+				if vals == nil {
+					vals = map[string][]byte{}
+				}
```
turkenh, line 155: These lines seem to be duplicates for both cases and could be combined.

pkg/types/builder.go
```diff
@@ -155,19 +155,25 @@ func (g *Builder) buildResource(res *schema.Resource, cfg *config.Resource, tfPa
 			sfx := "SecretRef"
 			cfg.Sensitive.AddFieldPath(fieldPathWithWildcard(tfPaths), "spec.forProvider."+fieldPathWithWildcard(xpPaths)+sfx)
 			// todo(turkenh): do we need to support other field types as sensitive?
-			if fieldType.String() != "string" && fieldType.String() != "*string" {
-				return nil, nil, fmt.Errorf(`got type %q for field %q, only types "string" and "*string" supported as sensitive`, fieldType.String(), fieldNameCamel)
+			if fieldType.String() != "string" && fieldType.String() != "*string" && fieldType.String() != "[]string" &&
+				fieldType.String() != "[]*string" {
+				return nil, nil, fmt.Errorf(`got type %q for field %q, only types "string", "*string", []string and []*string supported as sensitive`, fieldType.String(), fieldNameCamel)
```
turkenh, line 160: I remember @muvaf was [asking for](https://github.com/crossplane/terrajet/issues/100#issuecomment-974136406) `map[string]*string` support as well. 
Did you consider including that into this PR? Otherwise, I would suggest to create an issue for that while closing the original one with this PR.

pkg/resource/sensitive.go
```diff
@@ -181,21 +197,47 @@ func GetSensitiveParameters(ctx context.Context, client SecretClient, from runti
...
+					return err
+				}
+				if err = pavedTF.SetValue(expTF, sensitives); err != nil {
+					return errors.Wrapf(err, "cannot set string as terraform attribute for fieldpath %q", tfPath)
+				}
+			case map[string]interface{}:
+				sel := &v1.SecretKeySelector{}
+				if err = pavedJSON.GetValueInto(expandedJSONPath, sel); err != nil {
+					return errors.Wrapf(err, "cannot get SecretKeySelector from xp resource for fieldpath %q", expandedJSONPath)
+				}
```
turkenh, line 226: Please consider moving duplicate lines out.

pkg/resource/sensitive.go
```diff
@@ -181,21 +197,47 @@ func GetSensitiveParameters(ctx context.Context, client SecretClient, from runti
...
+				if err != nil {
+					return errors.Wrapf(err, "cannot get secret value for %v", sel)
+				}
+				expTF, err := expandedTFPath(expandedJSONPath, mapping)
+				if err != nil {
+					return err
+				}
+				if err = pavedTF.SetString(expTF, string(sensitive)); err != nil {
+					return errors.Wrapf(err, "cannot set string as terraform attribute for fieldpath %q", tfPath)
+				}
```
turkenh, line 237: Please consider moving duplicate lines out. (I think we can use `pavedTF.SetValue` for both.)

pkg/resource/sensitive_test.go
```diff
@@ -82,10 +82,6 @@ var (
 )
 
 func TestGetConnectionDetails(t *testing.T) {
-	testInput := map[string]interface{}{}
-	if err := json.JSParser.Unmarshal(testData, &testInput); err != nil {
-		t.Fatalf("cannot unmarshall test data: %v", err)
-	}
```
turkenh, line 88: 👍

turkenh (commented):
Thanks a lot @sergenyalcin for dealing with this.

PR is looking pretty good, left couple of comments mostly asking for some code cleanup.

pkg/types/builder.go
sergenyalcin: @turkenh  Thank you for this comment. As we discusses offline, providing the `map[string]*string` in a separate PR will be easier in terms of reviewing, testing and merging them.

I will open a separate PR and it will fixes the following issue (this issue is about that map[string]*string support): https://github.com/crossplane/terrajet/issues/208

pkg/resource/sensitive.go
sergenyalcin: Here, for the first case, a `[]v1.SecretKeySelector{}` (array of SecretKeySelector) and for the second case a `v1.SecretKeySelector{}` is used. Because of difference of types, I am not sure that, we can move these lines easily to out. If I am missing a point, could you please clarify?

turkenh: Ok, just had another look and you're right!

turkenh (approved):
Great work, thanks @sergenyalcin 👍

PR #198, opened 2022-01-13
Tagging Convention

sergenyalcin:

Fixes https://github.com/crossplane-contrib/provider-jet-aws/issues/136

This PR provides support for tagging convention for all providers in a generic way. By this implementation, custom initializers can be set.

A new configuration field, `Initializers`, was added. The type of this field is `[]InitializerFn`:

```
// InitializerFn returns the Initializer with a client.
type InitializerFn func(client client.Client) managed.Initializer
```

The field that is expected to be set has been implemented configurable on the basis of resource.

If the `Initializers` field is not set in the resource configuration, the initializers will not be generated in controllers in `Setup` function.

In three providers (aws, azure and gcp), a defaulting is not implemented for tag fields. Some statistics were collected to decide the topic:

AWS: 761 Generated Resource / 398 Resources has not "tags" field / 1 resource's "tags" field has different type (not map[string]*string)

Azure: 645 Generated Resource / 401 Resources has not "tags" field / 2 resources' "tags" field has different type (not map[string]*string)

Google: 436 Generated Resource / 376 Resources has not "labels" field / 2 resources' "labels" field has different type (not map[string]*string)

I have:

This PR was tested in provider-aws by using the S3-Bucket resource. Firstly, the commit id of this PR was used to consume the latest changes. And a custom `ExternalTagsFieldName` configuration was added for this resource. Then create, update and delete operations were successfully tested.

### DIFF
--- a/pkg/pipeline/templates/controller.go.tmpl
+++ b/pkg/pipeline/templates/controller.go.tmpl
@@ -26,12 +26,15 @@ import (
 // Setup adds a controller that reconciles {{ .CRD.Kind }} managed resources.
 func Setup(mgr ctrl.Manager, l logging.Logger, rl workqueue.RateLimiter, s terraform.SetupFn, ws *terraform.WorkspaceStore, cfg *tjconfig.Provider, concurrency int) error {
 	name := managed.ControllerName({{ .TypePackageAlias }}{{ .CRD.Kind }}_GroupVersionKind.String())
-	{{- if .Initializers }}
 	var initializers managed.InitializerChain
-	for _, i := range cfg.Resources["{{ .ResourceType }}"].Initializers {
+	{{- if .Initializers }}
+	for _, i := range cfg.Resources["{{ .ResourceType }}"].InitializerFns {
 	    initializers = append(initializers,i(mgr.GetClient()))
-    }
-    {{- end}}
+	}
+	{{- end}}
+	{{- if not .DisableNameInitializer }}
+	initializers = append(initializers, managed.NewNameAsExternalName(mgr.GetClient()))
+	{{- end}}
 	r := managed.NewReconciler(mgr,
 		xpresource.ManagedKind({{ .TypePackageAlias }}{{ .CRD.Kind }}_GroupVersionKind),
 		managed.WithExternalConnecter(tjcontroller.NewConnector(mgr.GetClient(), ws, s, cfg.Resources["{{ .ResourceType }}"],
@@ -43,13 +46,7 @@ func Setup(mgr ctrl.Manager, l logging.Logger, rl workqueue.RateLimiter, s terra
 		managed.WithRecorder(event.NewAPIRecorder(mgr.GetEventRecorderFor(name))),
 		managed.WithFinalizer(terraform.NewWorkspaceFinalizer(ws, xpresource.NewAPIFinalizer(mgr.GetClient(), managed.FinalizerName))),
 		managed.WithTimeout(3*time.Minute),
-		{{- if .Initializers }}
 		managed.WithInitializers(initializers),
-		{{else}}
-		{{- if .DisableNameInitializer }}
-		managed.WithInitializers(),
-		{{- end}}
-		{{- end}}
 		)
 
 	return ctrl.NewControllerManagedBy(mgr).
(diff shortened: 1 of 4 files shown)


pkg/pipeline/templates/terraformed.go.tmpl
```diff
@@ -92,3 +92,25 @@ func (tr *{{ .CRD.Kind }}) LateInitialize(attrs []byte) (bool, error) {
 func (tr *{{ .CRD.Kind }}) GetTerraformSchemaVersion() int {
     return {{ .Terraform.SchemaVersion }}
 }
+
+{{- if .ExternalTagsFieldName }}
+// SetExternalTags for this {{ .CRD.Kind }}
+func (tr *{{ .CRD.Kind }}) SetExternalTags() {
+	if tr.Spec.ForProvider.{{ .ExternalTagsFieldName }} == nil{
+		tr.Spec.ForProvider.{{ .ExternalTagsFieldName }} = map[string]*string{}
```
muvaf, line 100: I suspect that the type won't always be `map[string]*string{}`, which if changed would affect the whole function.

What are your thoughts about having the function implemented in provider repositories and refer to it as initializer function in the template? Provider authors can use whatever logic they'd like to and also possibly add more initializers if they need.

sergenyalcin: Hi @muvaf thanks for comment. Yes you are right that these fields are not always of a `map[string]*string{}` type. There are a few statistics about this in the description section of the PR. According to these data, it is very rare situation. However, it is possible that this situation may be seen more frequently in other providers. Therefore, it may be reasonable for us to handle this.

For the second paragraph of your comment:

https://github.com/crossplane-contrib/provider-jet-azure/blob/e04776f600c35b1c26e389937593b90fc6fbded1/config/kubernetes/config.go#L58

As you know, we have GetExternalNameFn field that is a function type. This field was set in providers side (by using the functions that are implemented in provider side). Do you mean something like this when the relevant function is implemented in providers?

muvaf: > According to these data, it is very rare situation.

Thanks for running these stats! I think accepting initializer will be useful for other stuff as well, so we can have the implementation you got here as the default, similar to other common defaults [here](https://github.com/crossplane/terrajet/blob/main/pkg/config/defaults.go#L28).

> Do you mean something like this when the relevant function is implemented in providers?

Yes, we can possibly accept an [`InitializerChain`](https://github.com/crossplane/crossplane-runtime/blob/295de47/pkg/reconciler/managed/reconciler.go#L143) and `DefaultResourceFn` can set it to the default set.

sergenyalcin: I updated the PR according to your comments. The implementations of setting tag fields are moved to provider repositories. Also, with last changes, this PR transformed from a tagging initializer PR to a generic initializer PR.

I am putting the provider PR links here to show the all process:

https://github.com/crossplane-contrib/provider-jet-azure/pull/130
https://github.com/crossplane-contrib/provider-jet-aws/pull/151
https://github.com/crossplane-contrib/provider-jet-gcp/pull/37

For `Initializers` field in resource configuration struct, `InitializerChain` type was not used directly to pass the `client.Client` in controllers. Instead of this, an `InitializerFn` type was used. This function type gets a `client.Client` as a parameter and returns an `managed.Initializer`. In controllers, before setting the `ReconcilerOptions`, the `Initializers` (configuration field) items is iterated and the client passes to the all `managed.Initializer`s.

pkg/pipeline/templates/controller.go.tmpl
```diff
@@ -37,9 +43,13 @@ func Setup(mgr ctrl.Manager, l logging.Logger, rl workqueue.RateLimiter, s terra
 		managed.WithRecorder(event.NewAPIRecorder(mgr.GetEventRecorderFor(name))),
 		managed.WithFinalizer(terraform.NewWorkspaceFinalizer(ws, xpresource.NewAPIFinalizer(mgr.GetClient(), managed.FinalizerName))),
 		managed.WithTimeout(3*time.Minute),
+		{{- if .Initializers }}
+		managed.WithInitializers(initializers),
+		{{else}}
 		{{- if .DisableNameInitializer }}
 		managed.WithInitializers(),
 		{{- end}}
+		{{- end}}
```
sergenyalcin, line 52: @muvaf We may discuss the possible issues that can occur the new logic about adding the WithInitializers option.

muvaf (approved):
LGTM! I've left small comments and as we synced offline, having a common implementation here and using it as tag initializer after configuring would be a nice UX and also inline with how we do defaulting in Terrajet in general. Thanks Sergen!

PR #199, opened 2022-01-14
Conversion: Default to string as element type

turkenh:

There could be some collection types which do not set types of their elements explicitly and we found out assuming `String` in that case is the right way to go. We have implemented this in [type builder](https://github.com/crossplane/terrajet/pull/182) but didn't follow same in conversion for resources using v1 sdk.

This PR fixes the same issue in conversion package.

I have:

With [provider rancher](https://registry.terraform.io/providers/rancher/rancher2/latest/docs/resources/catalog) which uses v1 sdk.

### DIFF
--- a/pkg/types/conversion/conversion.go
+++ b/pkg/types/conversion/conversion.go
@@ -85,7 +85,9 @@ func toV2Schema(v1sch *schema.Schema) *schemav2.Schema {
 		case *schema.Resource:
 			v2sch.Elem = toV2Resource(et)
 		default:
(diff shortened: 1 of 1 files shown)


ulucinar (approved):
Thanks @turkenh. Have we yet actually encountered this panic somewhere?

turkenh:
> Thanks @turkenh. Have we yet actually encountered this panic somewhere?

Yes, @ytsarev reported with with https://github.com/AbsaOSS/provider-jet-rancher

ulucinar:
Thanks for linking the issue.

PR #205, opened 2022-01-23
Terrajet provider generation documentation fixes

ytsarev:

* Fix broken link to Dockerfile
* Provide inline link for to terraform registry for clarity around `USE
  PROVIDER` button
* Extend sdkv1-v2 conversion documentation with `cmd/provider/main.go`

It's a documentation fix indirectly tested by the creation of https://github.com/AbsaOSS/provider-jet-rancher

### DIFF
--- a/docs/generating-a-provider.md
+++ b/docs/generating-a-provider.md
@@ -20,7 +20,7 @@ be quite similar for any other Terraform provider.
         export ProviderNameUpper=GitHub
         ```
 
-    2. Run the `./hack/prepare.sh` script from repo root to prepare the repo, e.g., to 
+    2. Run the `./hack/prepare.sh` script from repo root to prepare the repo, e.g., to
        replace all occurrences of `template` with your provider name:
 
         ```bash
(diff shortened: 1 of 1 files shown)


docs/generating-a-provider.md
```diff
@@ -87,6 +87,22 @@ be quite similar for any other Terraform provider.
...
+       ```go
+       import (
+           "github.com/crossplane/terrajet/pkg/types/conversion"
+       )
+
+       func main() {
+           ...
+
+           resourceMap := conversion.GetV2ResourceMap(tf.Provider())
+           kingpin.FatalIfError(controller.Setup(mgr, log, rl, setup, ws, pconfig.GetProvider(resourceMap), 1), "Cannot setup Template controllers")
```
turkenh, line 101: should we also include how `pconfig` imported?

ytsarev: @turkenh `pconfig` is already imported in the template https://github.com/crossplane-contrib/provider-jet-template/blob/main/cmd/provider/main.go#L32 .

Meanwhile `conversion` is not until you comment out https://github.com/crossplane-contrib/provider-jet-template/blob/main/cmd/provider/main.go#L77

So I think we need to highlight only `conversion` import here

turkenh: Alright, sounds good then.

turkenh (approved):
Thank you @ytsarev!

PR #219, opened 2022-02-02
Add a new section about Initializers to configuring resource doc

### LINKED ISSUE #204: Describing tagging configuration in documentation

In this [resource configuration document], we need to mention the new tagging configuration. 

Related changes: 

https://github.com/crossplane/terrajet/pull/198
https://github.com/crossplane-contrib/provider-jet-azure/pull/130
https://github.com/crossplane-contrib/provider-jet-aws/pull/151
https://github.com/crossplane-contrib/provider-jet-gcp/pull/37

[resource configuration document]: https://github.com/crossplane/terrajet/blob/main/docs/configuring-a-resource.md

sergenyalcin:

Fixes #204 

This PR adds a new section to configuring resource doc about Initializers. An example was added to the doc about tagging Initializers.

I have:

### DIFF
--- a/docs/configuring-a-resource.md
+++ b/docs/configuring-a-resource.md
@@ -11,6 +11,8 @@ documentation of the resource:
 - [Cross Resource Referencing]
 - [Additional Sensitive Fields and Custom Connection Details]
 - [Late Initialization Behavior]
+- [Overriding Terraform Resource Schema]
+- [Initializers]
 
 ### External Name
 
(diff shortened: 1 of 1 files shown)


docs/configuring-a-resource.md
```diff
@@ -557,13 +559,66 @@ p.AddResourceConfigurator("aws_autoscaling_group", func(r *config.Resource) {
...
+can be used as an initializer. The custom logic of the initializer will be implemented in `Initialize`
+function.
+
 [comment]: <> (References)
 
 [Terrajet]: https://github.com/crossplane/terrajet
 [External name]: #external-name
 [Cross Resource Referencing]: #cross-resource-referencing
 [Additional Sensitive Fields and Custom Connection Details]: #additional-sensitive-fields-and-custom-connection-details
-[Late Initialization Behavior]: #late-initialization-behavior
```
sergenyalcin, line 566: This reference is wrong, so the link does not work in document. Fixed with this PR.

docs/configuring-a-resource.md
```diff
@@ -11,6 +11,8 @@ documentation of the resource:
 - [Cross Resource Referencing]
 - [Additional Sensitive Fields and Custom Connection Details]
 - [Late Initialization Behavior]
+- [Overriding Terraform Resource Schema]
```
sergenyalcin, line 14: This section has not a reference. So I added with this PR. Please also check change in the line 621

docs/configuring-a-resource.md
```diff
@@ -557,13 +559,66 @@ p.AddResourceConfigurator("aws_autoscaling_group", func(r *config.Resource) {
 })
 ```
 
+### Initializers
+
+Initializers establish ownership of the supplied Managed resources in Crossplane.
+This typically involves the operations that are run before calling any
+ExternalClient methods.
```
muvaf, line 566: Seems like the definition is taken from runtime but I think it's wrong; there is no ownership operation done in initializer. I think scoping it as things that are done before the reconciliation begins would be more appropriate. Changing the comment in runtime as well would be good, too.

docs/configuring-a-resource.md
```diff
@@ -557,13 +559,66 @@ p.AddResourceConfigurator("aws_autoscaling_group", func(r *config.Resource) {
...
+
+This configuration option will provide that setting custom initializers for per
```
muvaf, line 568: For the readers of this document, it's important to know that whether they need the tool in given section or not right away. So, I'd start with an example case that is easy to understand, define the thing and then go into details of how they can do it. Something like why -> what -> how.

docs/configuring-a-resource.md
```diff
@@ -557,13 +559,66 @@ p.AddResourceConfigurator("aws_autoscaling_group", func(r *config.Resource) {
...
+
+Initializer is an interface in [crossplane-runtime]:
+
+```go
+type Initializer interface {
+	Initialize(ctx context.Context, mg resource.Managed) error
+}
+```
+
+So, an interface must be passed to the related configuration field for adding initializers for a resource.
```
muvaf, line 584: It could be helpful to note that tagging is most probably the only use case they would be concerned as an initializer, really, so that they don't feel like it's a widely used mechanism.

docs/configuring-a-resource.md
```diff
@@ -557,13 +559,66 @@ p.AddResourceConfigurator("aws_autoscaling_group", func(r *config.Resource) {
...
+Let's check an example about how can this configuration be used.
+
+`tags` is very common schema field for aws resources. Also, setting [some external labels] is a very common
+use case in Crossplane. Therefore, an initializer that sets the crossplane related labels for resources
+can be added in config file of the related resource:
+
+```go
+// Tagger implements the Initialize function to set external tags
+type Tagger struct {
```
muvaf, line 593: Just a suggestion, you can possible refer to the common implementation in Terrajet as example and also something they can use right away if it works for them instead of implying that they need to write it on their own.

muvaf (approved):

PR #237, opened 2022-02-18
Fix reconciliation errors if a referenced secret is deleted

### LINKED ISSUE #216: Reconciliation errors if a referenced secret is deleted.
### What happened?

If a secret is deleted before a terrajet resource is deleted, the following warning will be displayed and the resource will get stuck: 

```Warning  CannotConnectToProvider  23s (x334 over 6m11s)  managed/postgresql.azure.jet.crossplane.io/v1alpha1, kind=server  cannot get a terraform workspace for resource: cannot create a new file producer: cannot get sensitive parameters: cannot get secret value for &{{myazurepgserver-admin-password all-in-one-selfservice} password}: cannot get secret data: Secret "myazurepgserver-admin-password" not found```

### How can we reproduce it?

- Create resource with a secret
- Delete referenced secret before deleting the resource

### PR DESCRIPTION
**Pull request by ezgidemirel**
Signed-off-by: ezgidemirel <ezgidemirel91@gmail.com>

Fixes #216 

I have:

I replaced terrajet version consumed in `provider-jet-azure` to this branch in my local environment and created a PostgreSqlDB resource. Once it's ready, I first deleted the k8s secret referenced by `administratorLoginPasswordSecretRef` field and then deleted the MR. 

### DIFF
--- a/pkg/resource/sensitive.go
+++ b/pkg/resource/sensitive.go
@@ -25,6 +25,7 @@ import (
 	v1 "github.com/crossplane/crossplane-runtime/apis/common/v1"
 	"github.com/crossplane/crossplane-runtime/pkg/fieldpath"
 	"github.com/crossplane/crossplane-runtime/pkg/reconciler/managed"
+	"github.com/crossplane/crossplane-runtime/pkg/resource"
 	"github.com/pkg/errors"
 	kerrors "k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/runtime"
@@ -212,8 +213,14 @@ func GetSensitiveParameters(ctx context.Context, client SecretClient, from runti
 					sensitives := make(map[string]interface{})
 					for key, value := range *sel {
 						sensitive, err = client.GetSecretValue(ctx, value)
-						if err != nil {
+						if resource.IgnoreNotFound(err) != nil {
 							return errors.Wrapf(err, errFmtCannotGetSecretValue, sel)
+						}
+
+						// If referenced k8s secret is deleted before the MR, we pass empty string for the sensitive
+						// field to be able to destroy the resource.
+						if kerrors.IsNotFound(err) {
+							sensitive = []byte("")
 						}
 						sensitives[key] = string(sensitive)
 					}
@@ -226,7 +233,7 @@ func GetSensitiveParameters(ctx context.Context, client SecretClient, from runti
 						return errors.Wrapf(err, errFmtCannotGetSecretKeySelector, expandedJSONPath)
 					}
 					sensitive, err = client.GetSecretValue(ctx, *sel)
-					if err != nil {
+					if resource.IgnoreNotFound(err) != nil {
 						return errors.Wrapf(err, errFmtCannotGetSecretValue, sel)
 					}
 					if err := setSensitiveParametersWithPaved(pavedTF, expandedJSONPath, tfPath, mapping, string(sensitive)); err != nil {
@@ -241,8 +248,14 @@ func GetSensitiveParameters(ctx context.Context, client SecretClient, from runti
 				var sensitives []interface{}
 				for _, s := range *sel {
 					sensitive, err = client.GetSecretValue(ctx, s)
-					if err != nil {
+					if resource.IgnoreNotFound(err) != nil {
 						return errors.Wrapf(err, errFmtCannotGetSecretValue, sel)
+					}
+
+					// If referenced k8s secret is deleted before the MR, we pass empty string for the sensitive
+					// field to be able to destroy the resource.
+					if kerrors.IsNotFound(err) {
+						sensitive = []byte("")
 					}
 					sensitives = append(sensitives, string(sensitive))
 				}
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: pkg/resource/sensitive.go [hunk 1]
```diff
@@ -241,7 +226,7 @@ func GetSensitiveParameters(ctx context.Context, client SecretClient, from runti
 					return errors.Wrapf(err, errFmtCannotGetSecretKeySelector, expandedJSONPath)
 				}
 				sensitive, err = client.GetSecretValue(ctx, *sel)
-				if err != nil {
+				if err != nil && !kerrors.IsNotFound(err) {
```
turkenh, line 229: We can use [resource.IgnoreNotFound(err)](https://github.com/crossplane/crossplane-runtime/blob/428b7c3903756bb0dcf5330f40298e1fa0c34301/pkg/resource/resource.go#L207) instead.

pkg/resource/sensitive.go
```diff
@@ -203,22 +202,6 @@ func GetSensitiveParameters(ctx context.Context, client SecretClient, from runti
 			}
 
 			switch v.(type) {
-			case map[string]interface{}:
```
turkenh, line 206: What would the behavior be in the case of `map[string]secretRef` added to generated schema but we are dropping support here?

pkg/resource/sensitive.go
```diff
@@ -227,7 +210,9 @@ func GetSensitiveParameters(ctx context.Context, client SecretClient, from runti
 				var sensitives []interface{}
 				for _, s := range *sel {
 					sensitive, err = client.GetSecretValue(ctx, s)
-					if err != nil {
+					if err != nil && kerrors.IsNotFound(err) {
+						sensitive = []byte("")
```
turkenh, line 214: So, we have tested/validated terraform does not complain about having sensitive as an empty secret? e.g. password validation

pkg/resource/sensitive.go
ezgidemirel: done

pkg/resource/sensitive.go
ezgidemirel: We were seeing following error message on the resource:
```
Warning  CannotConnectToProvider  9s (x25 over 23s)  managed/dbforpostgresql.azure.jet.crossplane.io/v1alpha2, kind=server  cannot get a terraform workspace for resource: cannot create a new file producer: cannot get sensitive parameters: cannot unmarshal value from JSON: json: cannot unmarshal string into Go value of type v1.SecretKeySelector
```
Thanks to @sergenyalcin , the issue is fixed now.

pkg/resource/sensitive.go
ezgidemirel: That's correct. I have tested the behavior with both terraform cli (after updating the state and main.tf files) and with provider-jet-azure after consuming my change.

pkg/resource/sensitive.go
```diff
@@ -212,7 +213,9 @@ func GetSensitiveParameters(ctx context.Context, client SecretClient, from runti
 					sensitives := make(map[string]interface{})
 					for key, value := range *sel {
 						sensitive, err = client.GetSecretValue(ctx, value)
-						if err != nil {
+						if err != nil && kerrors.IsNotFound(err) {
+							sensitive = []byte("")
+						} else if err != nil {
 							return errors.Wrapf(err, errFmtCannotGetSecretValue, sel)
 						}
```
turkenh, line 220: nitpick:

```suggestion
						if resource.IgnoreNotFound(err) != nil {
							return errors.Wrapf(err, errFmtCannotGetSecretValue, sel)
						} 
						if kerrors.IsNotFound(err) {
							sensitive = []byte("")
						}
```

pkg/resource/sensitive.go
```diff
@@ -241,7 +244,9 @@ func GetSensitiveParameters(ctx context.Context, client SecretClient, from runti
 				var sensitives []interface{}
 				for _, s := range *sel {
 					sensitive, err = client.GetSecretValue(ctx, s)
-					if err != nil {
+					if err != nil && kerrors.IsNotFound(err) {
```
turkenh, line 247: same nitpick as above for better readability.

pkg/resource/sensitive.go
```diff
@@ -212,7 +213,9 @@ func GetSensitiveParameters(ctx context.Context, client SecretClient, from runti
 					sensitives := make(map[string]interface{})
 					for key, value := range *sel {
 						sensitive, err = client.GetSecretValue(ctx, value)
-						if err != nil {
+						if err != nil && kerrors.IsNotFound(err) {
+							sensitive = []byte("")
```
turkenh, line 217: A note on why we need this would be helpful for further readers.

turkenh (approved):
Added some nitpicks, looking good otherwise!
Thank you @ezgidemirel!

pkg/resource/sensitive.go
ezgidemirel: done

pkg/resource/sensitive.go
ezgidemirel: done

pkg/resource/sensitive.go
ezgidemirel: done

PR #244, opened 2022-02-24
Support for configuring blocks with optional parameters in attributes-as-blocks mode

### LINKED ISSUE #232: It is not possible to configure blocks with optional parameters
### What happened?

It [looks like](https://stackoverflow.com/a/69080432/4124416) while configuring attributes as values, users have to provide all configuration including optional ones. However, since Terrajet json tags optional parameters as `omitempty` while generating schema, those fields dropped and not passed to Terraform, even they were provided as zero values.

Not sure if this affects all resources/providers or a specific subset (needs to be investigated)

Originally reported in [this](https://github.com/crossplane-contrib/provider-jet-aws/issues/157) provider-jet-aws ticket. 

### How can we reproduce it?

See https://github.com/crossplane-contrib/provider-jet-aws/issues/157

sergenyalcin:

Fixes #232 

With this PR, configuring blocks with optional parameters will be possible. In terraform schemas, this logic works with a field: `ConfigMode`. If the value of the `ConfigMode` field is `SchemaConfigModeAttr`, we have to provide values for all options. Please see the details: https://stackoverflow.com/questions/69079945/terraform-inappropriate-value-for-attribute-ingress-while-creating-sg/69080432#69080432

Reference document: https://www.terraform.io/language/attr-as-blocks

In this PR, a new check was added for tfTag. If the parent field is in the `SchemaConfigModeAttr`, the `omitempty` tag is deleted from tftags of subfields.

I have:

This PR was tested in provider-jet-aws side by provisioning, updating and deleting SecurityGroup resource. In tests the manifest was used:

```
apiVersion: ec2.aws.jet.crossplane.io/v1alpha2
kind: SecurityGroup
metadata:
    name: example-securitygroup
spec:
  forProvider:
    name: example-securitygroup-sg
    description: "i am a description"
    region: us-west-2
    vpcId: vpc-somevpc
    ingress:
    - fromPort: 6379
      toPort: 6379
      protocol: tcp
      description: "i am also a description"
      cidrBlocks:
      - 10.0.0.0/16
      ipv6CidrBlocks: []
      self: false
      prefixListIds: []
      securityGroups: []
    tags:
      managed-by: crossplane
  providerConfigRef:
    name: jet-aws-provider
```

Please see the original issue in jet-aws side: https://github.com/crossplane-contrib/provider-jet-aws/issues/157

### DIFF
--- a/pkg/types/builder.go
+++ b/pkg/types/builder.go
@@ -64,7 +64,7 @@ type Builder struct {
 
 // Build returns parameters and observation types built out of Terraform schema.
 func (g *Builder) Build(cfg *config.Resource) (Generated, error) {
-	fp, ap, err := g.buildResource(cfg.TerraformResource, cfg, nil, nil, cfg.Kind)
+	fp, ap, err := g.buildResource(cfg.TerraformResource, cfg, nil, nil, false, cfg.Kind)
 	return Generated{
 		Types:           g.genTypes,
 		Comments:        g.comments,
@@ -73,7 +73,7 @@ func (g *Builder) Build(cfg *config.Resource) (Generated, error) {
 	}, errors.Wrapf(err, "cannot build the Types")
 }
 
-func (g *Builder) buildResource(res *schema.Resource, cfg *config.Resource, tfPath []string, xpPath []string, names ...string) (*types.Named, *types.Named, error) { //nolint:gocyclo
+func (g *Builder) buildResource(res *schema.Resource, cfg *config.Resource, tfPath []string, xpPath []string, asBlocksMode bool, names ...string) (*types.Named, *types.Named, error) { //nolint:gocyclo
 	// NOTE(muvaf): There can be fields in the same CRD with same name but in
 	// different types. Since we generate the type using the field name, there
 	// can be collisions. In order to be able to generate unique names consistently,
@@ -197,6 +197,9 @@ func (g *Builder) buildResource(res *schema.Resource, cfg *config.Resource, tfPa
 			obsFields = append(obsFields, field)
 			obsTags = append(obsTags, fmt.Sprintf(`json:"%s" tf:"%s"`, jsonTag, tfTag))
 		default:
+			if asBlocksMode {
+				tfTag = strings.TrimSuffix(tfTag, ",omitempty")
+			}
 			if sch.Optional {
 				paramTags = append(paramTags, fmt.Sprintf(`json:"%s" tf:"%s"`, jsonTag, tfTag))
 			} else {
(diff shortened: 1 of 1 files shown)


muvaf:
@sergenyalcin Thank you for that impact analysis! I think your proposal looks promising but it'd also mean that we can't move to CLI schema representation in the builder.

I looked at the reverse conversion function [here](https://github.com/hashicorp/terraform-plugin-sdk/blob/7e0a333644f1971a936995677b7a106140a0659f/helper/schema/core_schema.go#L90) and it seems like TF CLI does some things depending on the config mode. Have you had the chance to look at that to see if there is anything we can derive the config mode?

sergenyalcin:
> I looked at the reverse conversion function [here](https://github.com/hashicorp/terraform-plugin-sdk/blob/7e0a333644f1971a936995677b7a106140a0659f/helper/schema/core_schema.go#L90) and it seems like TF CLI does some things depending on the config mode. Have you had the chance to look at that to see if there is anything we can derive the config mode?

I checked this code block. As you said that in this function, there are some things depending on the `ConfigMode` values of fields. So, at first glance, it seems that doing the work done here in reverse may enable us to reach the relevant `ConfigMode` value.

When we check this code block in details, when the value of `ConfigMode` value is `SchemaConfigModeAttr` (this is the case we are trying to resolve), then the `coreConfigSchemaAttribute()` is calling. However, this function is calling for another cases: 
- https://github.com/hashicorp/terraform-plugin-sdk/blob/7e0a333644f1971a936995677b7a106140a0659f/helper/schema/core_schema.go#L86
- https://github.com/hashicorp/terraform-plugin-sdk/blob/7e0a333644f1971a936995677b7a106140a0659f/helper/schema/core_schema.go#L99

In other words, we see that the work done according to the `ConfigMode` value is not only done specifically for this value, but also in certain other cases. Therefore, trying to reach the `ConfigMode` value by running the work done here backwards may lead us to wrong results in some cases. 

For example, if the schema element is a map and the schema's Elem is a Resource object, the same function is called. So I'm not sure it's very possible to derive ConfigMode from the results in the generated schema. Please write if I'm missing a point.

> @sergenyalcin Thank you for that impact analysis! I think your proposal looks promising but it'd also mean that we can't move to CLI schema representation in the builder.

I got your point. So we may need to consider another configuration option to handle this problem. For example, while doing manuel configuration, the relevant fieldPath value can be passed for a configuration option, and we can check this fieldPath in builder.go instead of checking the ConfigMode value.

turkenh:
@sergenyalcin @muvaf just opened a fix/change setting `ConfigMode` during cli schema to sdk schema conversion. I've tested using existing provider-aws resources on main and looking good.

sergenyalcin:
@turkenh Since this PR https://github.com/crossplane/terrajet/pull/250 was merged, I think we may consider to review and  also merge this.

pkg/types/builder.go
```diff
@@ -269,9 +272,13 @@ func (g *Builder) buildSchema(sch *schema.Schema, cfg *config.Resource, tfPath [
...
 			}
 		case *schema.Resource:
+			var asBlocksMode bool
 			// TODO(muvaf): We skip the other type once we choose one of param
 			// or obs types. This might cause some fields to be completely omitted.
-			paramType, obsType, err := g.buildResource(et, cfg, tfPath, xpPath, names...)
+			if sch.ConfigMode == schema.SchemaConfigModeAttr {
+				asBlocksMode = true
+			}
+			paramType, obsType, err := g.buildResource(et, cfg, tfPath, xpPath, asBlocksMode, names...)
```
turkenh, line 281: The number of parameters that this method takes is annoying but given that we would refactor this part soon that should be fine for now.

turkenh (approved):

PR #255, opened 2022-03-07
Fix sensitive path mapping

### LINKED ISSUE #251: Inconsistent casing of 'JSON' when generating new struct with a sensitive field
### What happened?
I have a sensitive field called `data_json` in the upstream terraform resource.

When running code generation, this is converted into a sensitive field like so:
```golang
type SecretParameters struct {

	// JSON-encoded secret data to write.
	// +kubebuilder:validation:Required
	DataJSONSecretRef v1.SecretKeySelector `json:"dataJsonSecretRef" tf:"-"`
}
```

This creates a GetConnectionDetailsMapping like this:
```
// GetConnectionDetailsMapping for this Secret
func (tr *Secret) GetConnectionDetailsMapping() map[string]string {
	return map[string]string{"data": "status.atProvider.data", "data_json": "spec.forProvider.dataJSONSecretRef"}
}
```

This results in the error:
```bash
        The argument "data_json" is required, but no definition was found.
```

The  error clears when changing the mapping to `spec.forProvider.dataJsonSecretRef`.

### How can we reproduce it?
It would most likely affect any field which contained `_json` in the name

turkenh:

We incorrectly used `fieldName.LowerCamel` in sensitive field mapping which takes acronyms [here](https://github.com/crossplane/terrajet/blob/main/pkg/types/name/name.go#L104) into account. However, we should use `fieldName.LowerCamelComputed` instead since we don't use acronyms with fields names.

Fixes #251

I have:

Using provider-jet-vault, run make generate.

### DIFF
--- a/pkg/types/builder.go
+++ b/pkg/types/builder.go
@@ -119,7 +119,7 @@ func (g *Builder) buildResource(res *schema.Resource, cfg *config.Resource, tfPa
 		// Terraform paths, e.g. { "lifecycle_rule", "*", "transition", "*", "days" } for https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/s3_bucket#lifecycle_rule
 		tfPaths := append(tfPath, fieldName.Snake)
 		// Crossplane paths, e.g. {"lifecycleRule", "*", "transition", "*", "days"}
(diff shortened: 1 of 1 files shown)


sergenyalcin (approved):
LGTM!

PR #263, opened 2022-03-15
Allow late-initialization before a Terraform apply

ulucinar:

Due to an optimization related with the readiness of `Terraformed` resources, in certain circumstances, we never have a chance to do a late initialization before we run a Terraform apply. However, as revealed in https://github.com/crossplane-contrib/provider-jet-azure/issues/139, for certain resources and their configurations, it helps to do an early late-initialize before a Terraform plan. If certain optional fields are not set, a Terraform plan fails because we use the `prevent_destroy` lifecycle meta-argument. 

Categorically, we can also claim that these are upstream Terraform issues (e.g., one is [here](https://github.com/hashicorp/terraform-provider-azurerm/issues/8297)) but with this PR, we can both solve such cases and also preserve the readiness optimization we already have. For some context, please refer to https://github.com/crossplane-contrib/provider-jet-azure/issues/139 and please also see https://github.com/crossplane-contrib/provider-jet-azure/issues/139#issuecomment-1067630912. 

I have:

Tested via consuming this PR in `provider-jet-azure` in the context of https://github.com/crossplane-contrib/provider-jet-azure/issues/139. 

### DIFF
--- a/pkg/controller/external.go
+++ b/pkg/controller/external.go
@@ -162,23 +162,26 @@ func (e *external) Observe(ctx context.Context, mg xpresource.Managed) (managed.
 		return managed.ExternalObservation{}, errors.Wrap(err, "cannot get connection details")
 	}
 
+	lateInitedParams, err := tr.LateInitialize(res.State.GetAttributes())
+	if err != nil {
+		return managed.ExternalObservation{}, errors.Wrap(err, "cannot late initialize parameters")
+	}
+	lateInitialized := annotationsUpdated || lateInitedParams
+	resourceReady := tr.GetCondition(xpv1.TypeReady).Equal(xpv1.Available())
 	// We try to mark the resource ready before the spec late initialization
 	// and a call for "Plan" because those operations are costly, i.e. late-init
 	// causes a spec update which defers status update to the next reconcile and
-	// "Plan" takes a few seconds.
-	if !tr.GetCondition(xpv1.TypeReady).Equal(xpv1.Available()) {
+	// "Plan" takes a few seconds. We assume that a successful state refresh
+	// with state info available for the remote infrastructure in the output
+	// terraform.tfstate indicates readiness.
+	if !resourceReady || lateInitialized {
 		tr.SetConditions(xpv1.Available())
 		return managed.ExternalObservation{
 			ResourceExists:          true,
 			ResourceUpToDate:        true,
 			ConnectionDetails:       conn,
-			ResourceLateInitialized: annotationsUpdated,
+			ResourceLateInitialized: lateInitialized && resourceReady,
 		}, nil
-	}
-
-	lateInitedParams, err := tr.LateInitialize(res.State.GetAttributes())
-	if err != nil {
-		return managed.ExternalObservation{}, errors.Wrap(err, "cannot late initialize parameters")
 	}
 
 	plan, err := e.workspace.Plan(ctx)
@@ -189,7 +192,7 @@ func (e *external) Observe(ctx context.Context, mg xpresource.Managed) (managed.
 	return managed.ExternalObservation{
 		ResourceExists:          true,
 		ResourceUpToDate:        plan.UpToDate,
-		ResourceLateInitialized: annotationsUpdated || lateInitedParams,
+		ResourceLateInitialized: lateInitialized,
 		ConnectionDetails:       conn,
 	}, nil
 }
--- a/pkg/controller/external_test.go
+++ b/pkg/controller/external_test.go
@@ -21,12 +21,14 @@ import (
 	"testing"
 
 	xpv1 "github.com/crossplane/crossplane-runtime/apis/common/v1"
+	xpmeta "github.com/crossplane/crossplane-runtime/pkg/meta"
 	"github.com/crossplane/crossplane-runtime/pkg/reconciler/managed"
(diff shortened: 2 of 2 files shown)


ulucinar:
Please also see https://github.com/crossplane-contrib/provider-jet-azure/issues/139#issuecomment-1067630912.

pkg/controller/external.go
```diff
@@ -162,25 +162,26 @@ func (e *external) Observe(ctx context.Context, mg xpresource.Managed) (managed.
 		return managed.ExternalObservation{}, errors.Wrap(err, "cannot get connection details")
 	}
 
+	lateInitedParams, err := tr.LateInitialize(res.State.GetAttributes())
+	if err != nil {
+		return managed.ExternalObservation{}, errors.Wrap(err, "cannot late initialize parameters")
+	}
+	lateInitialized := annotationsUpdated || lateInitedParams
+	resourceReady := tr.GetCondition(xpv1.TypeReady).Equal(xpv1.Available())
 	// We try to mark the resource ready before the spec late initialization
```
turkenh, line 171: nit: looks like this comment needs to be updated, i.e. we no longer mark as ready before late init

pkg/controller/external.go
```diff
@@ -162,25 +162,26 @@ func (e *external) Observe(ctx context.Context, mg xpresource.Managed) (managed.
...
 	// and a call for "Plan" because those operations are costly, i.e. late-init
 	// causes a spec update which defers status update to the next reconcile and
 	// "Plan" takes a few seconds.
-	if !tr.GetCondition(xpv1.TypeReady).Equal(xpv1.Available()) {
+	if !resourceReady || lateInitialized {
 		tr.SetConditions(xpv1.Available())
```
turkenh, line 176: nit: I noticed that it was already the case before, but, confused to mark **the resource as ready when the resource is not ready** :) it would be great if you would mind clarifying this as well while updating the above comment.

turkenh (approved):
Should the PR title be `Allow late-initialization before a Terraform apply` ? 

It is `Allow late-initialization before a Terraform plan` and looking at this change, I see it was already before `plan` :)

pkg/controller/external.go
ulucinar: If I'm not mistaken, we still prioritize marking as ready over late-initialization, as `ExternalObservation.ResourceLateInitialized` will not be set to `true` if the resource does not have the `Ready` condition. And we prioritize a status update for initially marking the resource as ready over a spec update for actually late-initializing the resource. One of the goals of this change was to preserve that optimization as it affects the [time-to-readiness](https://github.com/crossplane/terrajet/issues/55#issuecomment-929494212) metric we defined for managed resources.

pkg/controller/external.go
ulucinar: Yes, that also confused me and caught my interest while preparing this PR. But as you mentioned this is already the case in main branch. I added a comment making this assumption explicit and maybe we can follow the implications of this assumption up with a separate issue.

pkg/controller/external.go
```diff
@@ -162,25 +162,28 @@ func (e *external) Observe(ctx context.Context, mg xpresource.Managed) (managed.
 		return managed.ExternalObservation{}, errors.Wrap(err, "cannot get connection details")
 	}
 
+	lateInitedParams, err := tr.LateInitialize(res.State.GetAttributes())
+	if err != nil {
+		return managed.ExternalObservation{}, errors.Wrap(err, "cannot late initialize parameters")
+	}
+	lateInitialized := annotationsUpdated || lateInitedParams
+	resourceReady := tr.GetCondition(xpv1.TypeReady).Equal(xpv1.Available())
```
muvaf, line 170: This is to make sure that if `unavailable -> available` transition is necessary, we will update the status hence no need to do any spec-related operations since they will be lost anyway. So `alreadyAvailable` or `markedAvailable` could be a less confusing than `resourceReady` because it doesn't represent the status in the cloud, it's only about whether the resource is marked as ready in etcd.

pkg/controller/external.go
```diff
@@ -162,25 +162,28 @@ func (e *external) Observe(ctx context.Context, mg xpresource.Managed) (managed.
...
+	// with state info available for the remote infrastructure in the output
+	// terraform.tfstate indicates readiness.
+	if !resourceReady || lateInitialized {
 		tr.SetConditions(xpv1.Available())
 		return managed.ExternalObservation{
 			ResourceExists:          true,
 			ResourceUpToDate:        true,
 			ConnectionDetails:       conn,
-			ResourceLateInitialized: annotationsUpdated,
+			ResourceLateInitialized: lateInitialized && resourceReady,
```
muvaf, line 183: Even if late-initialization is done, we should issue status update, i.e. `ResourceLateInitialized: false`, because readiness is more important than updating the spec. The only exception is the critical annotations we can't loose like external name.

muvaf (commented):
I think we should revert the change on line 175 where we used to issue spec update even if the resource is ready only if critical annotations need to be saved. @ulucinar I can open a PR if you'd like to.

pkg/controller/external.go
ulucinar: @muvaf thanks for the clarification for updating the critical annotations. I have opened this PR for further discussion:
https://github.com/crossplane/terrajet/pull/270

PR #267, opened 2022-03-21
Add shared gRPC mode to run the native provider

### LINKED ISSUE #261: Allow Terrajet-based providers to run in shared gRPC mode

We would like to have an option to prevent the Terraform CLI from forking the native provider plugin for each request, and instead communicate with a long running provider process via gRPC in the context of #38 and #233.

We can extent `provider-jet-template` and probably the configuration framework in Terrajet with options for configuring Terrajet-based providers to use a shared gRPC server.

ulucinar:

Fixes #261 
Adds a `config.Provider.SharedGRPC` provider configuration option to configure the shared gRPC native plugin mode.

I have:

- [x] Write tests for pkg/terraform/shared_grpc_server.go

Tested against [`provider-jet-template`](https://github.com/crossplane-contrib/provider-jet-template/pull/21) and with [`provider-jet-azure`](https://github.com/crossplane-contrib/provider-jet-azure/pull/166).


pkg/terraform/store.go
```diff
@@ -122,6 +141,9 @@ func (ws *WorkspaceStore) Workspace(ctx context.Context, c resource.SecretClient
 	}
 	l := ws.logger.WithValues("workspace", dir)
 	ws.mu.Lock()
+	if err := ws.startSharedServer(); err != nil {
```
muvaf, line 144: Another option could be to decouple the lifecycle of the provider server from the operations, i.e. maybe a struct different than `WorkspaceStore` and called directly in `cmd/main.go` of the providers and can restart the server if automatically if it crashes. The advantage would be that these parts would stay independent of the deployment model of the provider. Do you think that would be feasible?

pkg/terraform/shared_grpc_server.go
```diff
@@ -0,0 +1,95 @@
...
+			break
+		}
+		if err := cmd.Wait(); err != nil {
+			log.Info("Native Terraform provider process error", "error", err)
+			errCh <- err
+		}
+	}()
+
+	select {
+	case reattachConfig := <-reattachCh:
```
muvaf, line 86: What is reattach config exactly? If I assume it's configuration parameter about whether the server should be restarted or not, is there a case where we'd not like to do that?

pkg/terraform/shared_grpc_server.go
```diff
@@ -0,0 +1,95 @@
...
+	go func() {
+		defer close(errCh)
+		defer close(reattachCh)
+		defer func() {
+			ws.mu.Lock()
+			ws.reattachConfig = ""
+			ws.mu.Unlock()
+		}()
+		//#nosec G204 no user input
+		cmd := ws.executor.Command(ws.nativeProviderPath, ws.nativeProviderArgs...)
```
muvaf, line 59: A reminder for my comment [here](https://github.com/crossplane/terrajet/issues/261#issuecomment-1070827997). We should remove env var-based auth slowly since it won't work with this setup.

pkg/terraform/store.go
ulucinar: I think that should be feasible, if I understand you correctly, we could move anything related we do for the shared gRPC server to its own struct. But in order to keep things simpler I did not want to have a separate lifecycle management for the shared server. Ideally, I would have preferred a separate K8s deployment and a corresponding TCP service, which we can scale independently from the Crossplane provider. Or at least a sidecar container for which we could have separate liveness checks to have it managed by kubelet. I think coupling the lifecycle of the shared server with the workspace store strikes a good balance between simplicity & robustness and it's very similar to Terraform's production model, where the Terraform CLI automatically forks it before each call. With the proposed implementation, if we have some extra configuration for the workspacestore (namely the path of the binary plugin and the args to be passed to the native plugin process), with no separate logic to be implemented we have a robust, auto-starting server available. Do you think we would need a separate lifecycle for the shared server than the corresponding workspacestore? 

If you also agree with the initial demo deployment of this feature (where we had a separate K8s deployment for the shared server behind a K8s service), then I would propose to keep the lifecycle bound to workspacestore as of now until we have the necessary supporting features in the package manager to implement that. I like the simplicity we have now.

pkg/terraform/shared_grpc_server.go
ulucinar: The reattachment config tells the Terraform CLI not to fork the native provider itself but instead attach to an existing long running server, i.e., in our case, the shared server we fork. We configure the Terraform CLI's environment with this if we do not want it to fork the native plugin process itself. For this to work, we also need the native plugin (provider) to have support for this, i.e., it should be runnable outside of the Terraform CLI's environment. I have encountered binary plugins that refuse to be run if they are not in Terraform CLI's custom environment.

pkg/terraform/shared_grpc_server.go
muvaf: Actually I was meaning deprecating [this field](https://github.com/crossplane/terrajet/blob/1dfbdfcd7162105a67bd7ffb53b689f46a150450/pkg/terraform/store.go#L56) pointing that the configuration should be used instead of env.

pkg/terraform/store.go
muvaf: I have a slight preference for decoupling but I can see your point. Maybe then we can add an interface like `TerraformProviderRunner` with a `StartProvider` function? It would be called by `WorkspaceStore` just like here but it'd be satisfied by a `NopRunner` by default for example. Users can toggle what runner they want when they initialize the `WorkspaceStore` and the runners would be configured independently.

ulucinar: Done. I've added a `terraform. NativeProviderRunner` interface and the corresponding `providerRunner` member to the `terraform.WorkspaceStore` struct. By default a `terraoform. NoOpProviderRunner` is used.

pkg/terraform/shared_grpc_server.go
ulucinar: Oh. I see. I was considering `Setup.Env` as a general purpose configuration field that would allow us to inject arbitrary env. variables, not just those used for injecting authn env. variables. What do you think?

pkg/terraform/store.go
```diff
@@ -121,6 +133,13 @@ func (ws *WorkspaceStore) Workspace(ctx context.Context, c resource.SecretClient
 		return nil, errors.Wrap(err, "cannot write main tf file")
 	}
 	l := ws.logger.WithValues("workspace", dir)
+	reattachConfig, err := ws.providerRunner.StartSharedServer()
+	if err != nil {
+		return nil, err
+	}
+	if err := os.Setenv(envReattachConfig, reattachConfig); err != nil {
```
muvaf, line 140: There will be concurrent calls to this `Workspace` function, so setting an OS-wide env var may not work well here. If it's a required env var for every TF CLI invocation, maybe we can set it as env var to `*Workspace` object retrieved from the store map?

pkg/terraform/store.go
```diff
@@ -121,6 +133,13 @@ func (ws *WorkspaceStore) Workspace(ctx context.Context, c resource.SecretClient
 		return nil, errors.Wrap(err, "cannot write main tf file")
 	}
 	l := ws.logger.WithValues("workspace", dir)
+	reattachConfig, err := ws.providerRunner.StartSharedServer()
```
muvaf, line 136: I got to learn what this might return from our offline chat. `attachmentConfig` could be less specific so that others understand what it is immediately. Or maybe `terraformEnvConfig` that we can set directly.

pkg/terraform/provider_runner.go
```diff
@@ -0,0 +1,93 @@
...
+
+// NativeProviderRunner is the interface for running
+// Terraform native provider processes in the shared
+// gRPC server mode
+type NativeProviderRunner interface {
+	StartSharedServer() (string, error)
+}
+
+// NoOpProviderRunner is a no-op NativeProviderRunner
+type NoOpProviderRunner struct{}
```
muvaf, line 35: Having a `NewNoOpProviderRunner() NoOpProviderRunner` would make it more consistent I think.

muvaf (commented):
Looking pretty good, most of the comments are about styling. Thanks @ulucinar !

pkg/terraform/shared_grpc_server.go
```diff
@@ -0,0 +1,104 @@
...
+
+	"github.com/pkg/errors"
+)
+
+const (
+	// error messages
+	errNativeProviderPath = "native provider path is not configured"
+	errFmtTimeout         = "timed out after %v while waiting for the reattach configuration string"
+
+	envReattachConfig = "TF_REATTACH_PROVIDERS"
```
muvaf, line 32: An example value as a comment would definitely be helpful for future readers.

pkg/terraform/store.go
```diff
@@ -133,6 +150,7 @@ func (ws *WorkspaceStore) Workspace(ctx context.Context, c resource.SecretClient
 		return nil, errors.Wrap(err, "cannot stat init lock file")
 	}
 	w.env = ts.Env
+	w.env = append(w.env, fmt.Sprintf(fmtEnv, envReattachConfig, attachmentConfig))
```
muvaf, line 153:
```suggestion
	w.env = append(ts.Env, fmt.Sprintf(fmtEnv, envReattachConfig, attachmentConfig))
```
nitpick

pkg/terraform/store.go
```diff
@@ -66,14 +71,22 @@ func WithFs(fs afero.Fs) WorkspaceStoreOption {
...
-		logger:   l,
-		mu:       sync.Mutex{},
-		fs:       afero.Afero{Fs: afero.NewOsFs()},
-		executor: exec.New(),
+		store:          map[types.UID]*Workspace{},
+		logger:         l,
+		mu:             sync.Mutex{},
+		fs:             afero.Afero{Fs: afero.NewOsFs()},
+		executor:       exec.New(),
+		providerRunner: NewNoOpProviderRunner(),
```
muvaf, line 89: nitpick: since jet-template uses grpc by default, we can also use it by default, too, and document in the guide that you can use NoOp one if shared grpc doesn't work for you for some reason.

muvaf (approved):
LGTM! Thank you @ulucinar !

pkg/terraform/store.go
muvaf: Hmm I thought it was process-safe, i.e. process setting its own execution, hence the forked tf calls would use the set of env vars at the moment they were forked.

pkg/terraform/shared_grpc_server.go
muvaf: I'm not sure. It's only used for credentials and they have to go to provider, so there isn't a use case coming to my mind that you want to affect TF CLI's env vars.

pkg/terraform/provider_runner.go
```diff
@@ -0,0 +1,182 @@
...
+// A logger, native provider's path and command-line arguments to be
+// passed to it must have been properly configured.
+// Returns any errors encountered and the reattachment configuration for
+// the native provider.
+func (sr *SharedProvider) Start() (string, error) { //nolint:gocyclo
+	sr.mu.Lock()
+	defer sr.mu.Unlock()
+	if len(sr.nativeProviderPath) == 0 {
+		return "", errors.New(errNativeProviderPath)
+	}
```
muvaf, line 125: We can require this by taking it as direct argument on `NewSharedProvider`

pkg/terraform/store.go
ulucinar: I would prefer to have an initial set of providers in the wild for some time before making this the default. What do you think?

pkg/terraform/provider_runner.go
```diff
@@ -0,0 +1,170 @@
...
+
+// Start starts a shared gRPC server if not already running
+// A logger, native provider's path and command-line arguments to be
+// passed to it must have been properly configured.
+// Returns any errors encountered and the reattachment configuration for
+// the native provider.
+func (sr *SharedProvider) Start() (string, error) { //nolint:gocyclo
+	sr.mu.Lock()
+	defer sr.mu.Unlock()
+	log := sr.logger.WithValues("nativeProviderPath", sr.nativeProviderPath, "nativeProviderArgs", sr.nativeProviderArgs)
```
muvaf, line 114: nitpick: we can do this initialization of `log` in `NewSharedProvider` so that it's done only once.

PR #270, opened 2022-03-28
Prioritize critical resource annotation updates over status updates

ulucinar:

Recently with #263, we have deprioritized critical annotation updates wrt status updates. However, we would like to have them prioritized over all others.

I have:

Tested against `provider-jet-azure` by provisioning a resource with it.

### DIFF
--- a/pkg/controller/external.go
+++ b/pkg/controller/external.go
@@ -167,12 +167,18 @@ func (e *external) Observe(ctx context.Context, mg xpresource.Managed) (managed.
 		return managed.ExternalObservation{}, errors.Wrap(err, "cannot late initialize parameters")
 	}
 	markedAvailable := tr.GetCondition(xpv1.TypeReady).Equal(xpv1.Available())
-	// We try to mark the resource ready before the spec late initialization
-	// and a call for "Plan" because those operations are costly, i.e. late-init
-	// causes a spec update which defers status update to the next reconcile and
-	// "Plan" takes a few seconds. We assume that a successful state refresh
-	// with state info available for the remote infrastructure in the output
-	// terraform.tfstate indicates readiness.
+	// In the following switch block, before running a relatively costly
+	// Terraform apply and that may fail before critical annotations are
+	// updated, or late-initialized configuration is written to main.tf.json,
+	// we try to perform the following in the given order:
+	// 1. Update critical annotations if they have changed
+	// 2. Update status
+	// 3. Update spec with late-initialized fields
+	// We prioritize critical annotation updates most not to lose certain info
+	// (like the Cloud provider generated ID) before anything else. Then we
+	// prioritize status updates over late-initialization spec updates to
+	// mark the resource as available as soon as possible because a spec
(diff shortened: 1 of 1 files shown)


pkg/controller/external.go
```diff
@@ -192,7 +195,7 @@ func (e *external) Observe(ctx context.Context, mg xpresource.Managed) (managed.
 	return managed.ExternalObservation{
 		ResourceExists:          true,
 		ResourceUpToDate:        plan.UpToDate,
-		ResourceLateInitialized: lateInitialized,
+		ResourceLateInitialized: annotationsUpdated || lateInitedParams,
```
ulucinar, line 198: If I'm not missing anything, we can always set this to `False` because we will already have late-inited the managed resource in the above block (maybe after a status update iteration).

pkg/controller/external.go
```diff
@@ -166,21 +166,24 @@ func (e *external) Observe(ctx context.Context, mg xpresource.Managed) (managed.
...
+	markedAvailable := tr.GetCondition(xpv1.TypeReady).Equal(xpv1.Available())
 	// We try to mark the resource ready before the spec late initialization
 	// and a call for "Plan" because those operations are costly, i.e. late-init
 	// causes a spec update which defers status update to the next reconcile and
 	// "Plan" takes a few seconds. We assume that a successful state refresh
 	// with state info available for the remote infrastructure in the output
 	// terraform.tfstate indicates readiness.
-	if !resourceReady || lateInitialized {
+	if !markedAvailable || annotationsUpdated || lateInitedParams {
 		tr.SetConditions(xpv1.Available())
```
ulucinar, line 177: I also feel like we should be doing this as a function of the `plan` output, what do you think?

pkg/controller/external.go
```diff
@@ -166,21 +166,24 @@ func (e *external) Observe(ctx context.Context, mg xpresource.Managed) (managed.
...
-	resourceReady := tr.GetCondition(xpv1.TypeReady).Equal(xpv1.Available())
+	markedAvailable := tr.GetCondition(xpv1.TypeReady).Equal(xpv1.Available())
 	// We try to mark the resource ready before the spec late initialization
 	// and a call for "Plan" because those operations are costly, i.e. late-init
 	// causes a spec update which defers status update to the next reconcile and
 	// "Plan" takes a few seconds. We assume that a successful state refresh
 	// with state info available for the remote infrastructure in the output
 	// terraform.tfstate indicates readiness.
-	if !resourceReady || lateInitialized {
+	if !markedAvailable || annotationsUpdated || lateInitedParams {
```
muvaf, line 176: It took me quite a while to understand the reasoning of all the conditions in our offline chat and my old comment doesn't seem to be doing a good job there anymore. What do you think about a `switch` statement with cases for each of the conditions and whose `default:` is calling the plan & return? I think it'd end up in the same flow but just much easier to follow what's going on and why.

muvaf (approved):
LGTM! Thanks @ulucinar for bearing with me while I was understanding the flow!

PR #290, opened 2022-08-03
Add float64 to supported sensitve fields

Lax77:
Add support for float64 data type as sensitive field. We ran into an issue while upgrading provide-jet-datadog to latest release of [terraform-provider-datadog](https://github.com/DataDog/terraform-provider-datadog) as it introduced numeric sensitive field  [client_token_id](https://registry.terraform.io/providers/DataDog/datadog/latest/docs/resources/synthetics_test#nested-schema-for-options_listrum_settings) of type float64

Fixes # https://github.com/crossplane/terrajet/issues/287

I have:

- [ ] Read and followed Crossplane's [contribution process].

- `make generate` on a jet provider

- `make reviewable test` on a jet provider


Kerwood:
How is it going with this one ? Can we get this accepted ?

pkg/types/field.go
```diff
@@ -98,7 +99,7 @@ func NewSensitiveField(g *Builder, cfg *config.Resource, r *resource, sch *schem
 
 	f.TFTag = "-"
 	switch f.FieldType.String() {
-	case "string", "*string":
+	case "string", "*string", "float64", "*float64":
```
muvaf, line 102: Have you tested this change manually and confirmed it works? Here is where we take the value from secret and then write to the TF file but it converts it to `string` before doing so. So, I think we may end up having a `"231231241"` as value instead of `231231241` in the `main.tf.json`, which Terraform may accept but better check if that's the case. Otherwise, we need to find a way to discover the type and convert it to that type before calling `setSensitiveParametersWithPaved` function. cc @turkenh

Lax77: I have only tried upgrading [terraform-provider-datadog](https://github.com/DataDog/terraform-provider-datadog) to version v3.14 which seems to work. But did not try creating a resource with it that uses float64 for sensitive fields check. Will test it out and get back

PR #296, opened 2022-08-17
go.mod: update go-getter to the latest to avoid reported CVEs

muvaf:

v1 of `terraform-plugin-sdk` has reached EOL so we have to bump the dependency separately. Removal of v1 SDK completely is discussed [here](https://github.com/crossplane/terrajet/issues/294) but this PR should unblock folks.

Fixes https://github.com/crossplane/terrajet/issues/294

I have:

`make reviewable`


go.mod
```diff
@@ -2,20 +2,23 @@ module github.com/crossplane/terrajet
 
 go 1.17
 
+// Due to https://github.com/crossplane/terrajet/issues/294
+replace github.com/hashicorp/go-getter => github.com/hashicorp/go-getter v1.6.2
```
jbw976, line 6: do we have a replace directive because we do not directly depend on this package? 🤔

jbw976 (approved):

